PR #2, opened 2022-01-15
Improved data/query params

Sammyjo20:
## Changes
- Re-written the way formData and queryParams are processed
- formData and queryParams can now be added to the base connector


Sammyjo20 (commented):
Self review complete

PR #3, opened 2022-01-17
Added support for Guzzle handlers

Sammyjo20:
## Changes
- Added support for Guzzle handlers to customise the Guzzle request/response before it is sent out 🔥 (Caching coming soon 😉)
- Improved interceptors
- Added "boot" method to request/connector to easily add functionality

## Breaking changes
- Removed old interceptors pattern


Sammyjo20 (commented):
Self review complete

Issue #8, opened 2022-01-20
License


simonhamp:
I can see you've got the LICENSE file in `.github/`, but it might be worth setting it somewhere more public.

I think if it's in the root of the project, GitHub can pick it up and pop a link to it in the sidebar

<img width="377" alt="Screenshot 2022-01-20 at 17 48 56" src="https://user-images.githubusercontent.com/31628/150393937-cfe2117d-2ad2-48bf-996e-ed1ed04a6365.png">

Sammyjo20:
Thanks for spotting this! I've resolved this in #9

Issue #6, opened 2022-01-18
Merged URL not valid


alexcutts:
This is a great project - exactly what I needed to standardise our API connectivity as we use a lot of integrations. I realise this is a problem with the third party API I'm using not confirming to REST specifications but want to check if anything can be done.

We use an API which has the baseUrl format `api.xxx.com/v2`
Requests are then made to endpoints in the format `model=accounts&action=load`
This leads to an actual URL of `api.xxx.com/v2?model=accounts&action=load`

Saloon constructs the URL by combining the two parts, trimming the right forward slash (if present) and then inserting a forward slash - e.g `api.xxx.com/v2/?model=accounts&action=load` - this produces a 404 error as it is now referencing a non existent directory. 

I have tried using handlers but you can't modify the base URL once set (done in `ManagesGuzzle`).

The easy solution is to just use the full URL in the Request class but this takes away some of the simplicity that this package created.

I can't think of any other way of modifying the URL before the request is made - however an option on the SaloonConnector class could allow the behaviour to be configured at runtime per connector (or alternatively, an option in the global config to add slashes on a global basis). I'm happy to submit a pull request if this is something you'd be interested in adding

Sammyjo20:
Thanks for submitting this issue, really glad you like Saloon! This definitely sounds like a bit of a bug. Where I intended to always only have one trailing slash, it adds another. I believe a fix for this would be to not add any leading slash if it doesn’t exist already, but if one exists already - leave it in there. 

If you are happy to submit a PR for this, that would be great - if not I’ll try and add it in later today after work.

Sammyjo20:
Actually thinking about this more - if I remove the logic that fixes trailing/leading slashes - it will become a breaking change as people will assume Saloon will fix their leading/trailing slashes (a problem that I have had before)

Maybe Saloon could introduce a new trait like “DoNotIncludeTrailingSlash” or something?

Alternatively, I can remove the logic where I add a trailing slash in the ManagesGuzzle file in the base url. This would cause people’s APIs to break unless they put a leading slash in the request.

Sammyjo20:
Sorry for lots of comments! To be honest, I think a "DoNotIncludeTrailingSlash" would do the trick because that's the only thing that you can't control at the moment with how Saloon builds URLs. If I added that trait - you would be responsible to add the trailing slash but you could do the rest in the request endpoint.

Sammyjo20:
[...]

PR #9, opened 2022-01-20
Disable trailing slash

Sammyjo20:
## Changes

- Now if you don't provide an endpoint in the request, Saloon won't add a trailing slash to the base url (#6)
- Moved license file to root directory (#8)


Sammyjo20 (commented):
Self review complete

PR #12, opened 2022-01-22
Fix typo in readme

adil62:
Changed the below code 

From

`$response->getStatusFroMResponse(): int`

To

`$response->getStatusFromResponse(): int`


Sammyjo20:
Thank you! ❤️

Issue #11, opened 2022-01-22
Add casting to DTO


Wulfheart:
This package looks really nice. 
Currently I am using https://github.com/spatie/data-transfer-object to cast the json response to a class and vice versa. 
Do you know any good way how to integrate libraries like this into the `Request` classes?

Sammyjo20:
Thanks @Wulfheart!

Maybe if Saloon integrated [the macroable trait](https://github.com/spatie/macroable) by Spatie on the Response, you would be able to add your own method within a response interceptor?

Sammyjo20:
I have added the `Macroable` trait to the response so you can easily add your own methods to extend it.

Wulfheart:
Hey,
sorry for my late reply. I don't think that the `Macroable` trait is the right way forward with this issue.
For your consideration:
What about a method in the request like `responseToDto($response)`? I know that it is not in the response itself but way better than a macro in my opinion. However this would require the docblocks to be generic for the correct typehinting in the `$response->toDTO()`.

yurirnascimento:
Olá @Wulfheart,

I also use the Spatie/Data-Transfer-Object package, a way that this can be solved is directly in your DTO, for example:

## The DTO

```
use \Spatie\DataTransferObject\DataTransferObject;\
use \Illuminate\Http\Request;

class TestDTO extends DataTransferObject {

	public string $someValue;

	public static function fromRequest(Request $request){
		return new self([
			"someValue" => $request->input("someValue")
		]);
	}
	
	public static function fromAPIResponse(array $json) : self
	{
		return new self([
			"someValue" => data_get($json, "someValue")
		]);
	}

}
```

## Action that use TestDTO

```
class TestAction {
	
	public function __invoke(TestDTO $testDTO): TestModel
	{
		
		$test = TestModel($testDTO->all());
		$test->save();
		
		return $test;
	}

}
```

## Ways to use

```
// From Saloon Request
$request = new SomeSaloonRequest();
$dto = TestDTO::fromResponse($request->send());

// Or from Controller
public function store(Request $request, TestAction $testAction)
{
	return TestResource::make(
		$testAction(TestDTO::fromRequest($request))
	);
}
```

Using this way, in my opinion, it has a better separation, and the DTO can implement other ways to initialize itself.

Sorry for the offtopic, I just don't think Saloon would need to solve this.

Wulfheart:
Hey @yurirnascimento,

thanks for your thorough overview. 
[...]

PR #16, opened 2022-01-27
WIP: Prototype for custom responses

Wulfheart:
Related to #11.

This is a prototype and can be tried out at https://github.com/Wulfheart/SaloonWorkbench. It is necessary to have Saloon and SaloonWorkbench in the same directory.
In the end there have to be more tests and a `saloon:response` command. The biggest pain point right now is the lack of typehinting for the response on `$request->send()`. I think it can be solved with docblocks generics like described [here](https://phpstan.org/blog/generics-in-php-using-phpdocs) but I have no idea how to implement it. Any ideas regarding this?

### DIFF
--- a/src/Http/SaloonConnector.php
+++ b/src/Http/SaloonConnector.php
@@ -20,6 +20,21 @@ abstract class SaloonConnector implements SaloonConnectorInterface
         CollectsInterceptors;
 
     /**
+     * The response class.
+     *
+     * @var class-string<\Sammyjo20\Saloon\Http\SaloonResponse>|null
+     */
+    protected ?string $response = null;
+
+    /**
+     * Gets the response class.
+     * @return string|null
+     */
+    public function getResponseClass(): ?string {
+        return $this->response;
+    }
+
+    /**
      * Define anything to be added to the connector.
      *
      * @return void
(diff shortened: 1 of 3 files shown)


Sammyjo20:
Looks good so far. Why the need for a separate saloon workbench? I’ve been able to test everything in the tests up until now. Just curious!

Wulfheart:
Just wanted to make a project where I can test the enduser experience manually including the PHPStorm autocompletion.

yurirnascimento:
Hi @Wulfheart 

One way I saw Spatie getting "around" this, would be like this:

```php
class UserRequestResponse extends SaloonRequest
{

	protected ?string $response = UserResponse::class;

	public function send(MockClient $mockClient = null): UserResponse
	{
		return parent::send($mockClient);
	}
}
```

Using this in an earlier version of Data Transfer Object.

You can see this [here](https://github.com/spatie/data-transfer-object/tree/v2#working-with-collections)

Hope this can help.

yurirnascimento:
I'm implementing a test with docblocks generics but it seems that my PHPStorm is not reading correctly.

Sammyjo20:
I’m happy without the docblocks generics for now. It should list you with all the methods of SaloonResponse because it is type-hinted.

Sammyjo20:
I really appreciate this PR, but I think we now have a complete PR #17 almost ready, @Wulfheart I think it comes with everything we need, let me know if you would add anything.

Wulfheart:
Nice. 🎉

PR #20, opened 2022-01-28
Removed requirement for HasQueryParams trait

Sammyjo20:
This PR removes the `HasQueryParams` trait and will add the query params automatically if you pass in any query parameters.


Sammyjo20 (commented):
Self review complete

PR #21, opened 2022-01-29
Add default defineEndpoint in base SaloonRequest

ajthinking:
Just got started with this for a graphql integration, works really good 💫 
Not *100% sure* this is an improvement you would like so feel free to decline it 👍 

Since graphql request always targets a root like "www.acme.com/api/graphql" - no appended endpoint - would it make sense to allow to not explicitly repeating defineEndpoint method returning '' every time?
```php
    /**
     * Define the endpoint for the request.
     *
     * @return string
     */
    public function defineEndpoint(): string
    {
        return '';
    }
```    
This PR adds the above to the base class as a default.
   
Also let me know if you prefer keeping all tests in function block style. I just discovered that higher order thing 😄

### DIFF
--- a/src/Http/SaloonRequest.php
+++ b/src/Http/SaloonRequest.php
@@ -137,6 +137,16 @@ abstract class SaloonRequest implements SaloonRequestInterface
     }
 
     /**
+     * Define the endpoint for the request.
+     *
+     * @return string
+     */
+    public function defineEndpoint(): string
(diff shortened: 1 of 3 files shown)


Sammyjo20 (approved):
This is a great little improvement! 🙌 Glad it is working well for GraphQL

Thank you for the PR ❤️

PR #22, opened 2022-01-30
Now returning original error and tidied up response

Sammyjo20:
## Added
- New `getGuzzleException()` to SaloonResponse
- Previous exception to `SaloonRequestException`

## Changed
- Tidied up "Should Guess Status From Body" feature
- Removed `effectiveUri()` and `handlerStats()` from response
- Renamed `getSaloonRequestOptions` to `getRequestOptions`


Sammyjo20 (commented):
Self review complete

PR #25, opened 2022-02-02
Extended exceptions

Sammyjo20:
## Changes

- All Saloon's exceptions now extend "SaloonException"
- Added tests for response methods


Sammyjo20 (commented):
Self review complete

PR #28, opened 2022-02-04
New connector request API

Sammyjo20:
PR in progress

Todo: Laravel "auto discovery" trait
Todo: Add interface method for setLoadedConnector on the request
Todo: Make getRegisteredRequests public for people to use 
Todo: Add requestExists() method to check if a registered request exists


Sammyjo20 (commented):
Self review complete

PR #33, opened 2022-02-11
Update HasFormParams.php

ronnievisser:
bootHasBodyFeature was never called when using this trait since its not folowing the correct naming convention.

This change solved that issue and the data is attached correctly.


Sammyjo20:
@ronnievisser This is now in v0.7.1

PR #34, opened 2022-02-11
Fixed an issue where it would not check against history

### LINKED ISSUE #32: Fix AssertSentJson
I'm aware of an issue with AssertSentJson if there are multiple requests that have been sent using the same Saloon request but with different data.

How to reproduce: 
- Send two mock requests 
- Use the `AssertSentJson` method to check the data in the second request
- It will fail because it will pull out the first request and check the data against the first match.

Sammyjo20:
Fixes #32 and solves an issue where if you used a closure for `assertSent` it would only check the latest request.

### DIFF
--- a/src/Clients/BaseMockClient.php
+++ b/src/Clients/BaseMockClient.php
@@ -283,11 +283,9 @@ class BaseMockClient
      */
     public function assertSentJson(string $request, array $data): void
     {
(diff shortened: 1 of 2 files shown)


Sammyjo20 (commented):
Self review complete

PR #40, opened 2022-02-28
Added new 'CastsToDto' plugin

Sammyjo20:
This PR introduces a first-party way to cast directly to a DTO of your choice.

Firstly, add the "CastsToDto" trait to your request. After that, you need to add the "castToDto" method on the request. This should return your DTO with the data populated inside of it. 

```php
public function castToDto(SaloonResponse $response): object
{
    return ApiResponse::fromSaloon($response);
}
```

After your request has been sent, you can use the `dto` method on the response to retrieve your DTO.

```php
$response = UserRequest::make()->send();
$dto = $response->dto(); // Data\User
```


Sammyjo20 (commented):
Self review complete

PR #47, opened 2022-04-10
Added new method to send requests on connectors

Sammyjo20:
- Added new `send` and `request` methods to the connector
- Changed return type of cast
- Moved make back out from trait to improve readability


Sammyjo20 (commented):
Self review complete

PR #56, opened 2022-04-24
Feature | Asynchronous requests

Sammyjo20:
Introducing support for Guzzle's asynchronous requests in Saloon! 🎉

This PR introduces a new `sendAsync` method on the request/connector that you can use to return a [Guzzle Promise](https://github.com/guzzle/promises).

```php
$request = new GetUserRequest(userId: 1);
$promise = $request->sendAsync();

$promise->then(
    function (SaloonResponse $response) {
        // Handle successful response...
    },
    function (SaloonRequestException $exception) {
        $response = $exception->getSaloonResponse();

        // Handle unsuccessful response...
    },
)
```

### Important Note
Typically, Saloon will return a response with the error inside of the response, however with promises, if an error occurs - Saloon will return the response in the rejection chain instead. Should this be the case? Or should we only reserve rejection for exceptions like ConnectException and always return responses, even if errored in the resolve callback?

### DIFF
--- a/src/Http/SaloonConnector.php
+++ b/src/Http/SaloonConnector.php
@@ -2,6 +2,7 @@
 
 namespace Sammyjo20\Saloon\Http;
 
+use GuzzleHttp\Promise\PromiseInterface;
 use ReflectionClass;
 use Illuminate\Support\Str;
 use Illuminate\Support\Collection;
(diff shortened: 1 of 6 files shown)


Sammyjo20 (commented):
Self review complete

juse-less:
@Sammyjo20 Regarding your note, and the way you handle the `onRejection` - I've only ever seen `Exception`s _`return`ed_, not `throw`nd.
Curious as I am, however, I started digging more into Guzzles code.
I can't fully grasp the whole flow (at least not by browsing it on GitHub), but.. it's interesting if `throw`ing an `Exception`, like you do, [would create a `RejectionPromise`](https://github.com/guzzle/guzzle/blob/master/src/Client.php#L333), and, when `return`ing instead, [the `FulfilledPromise` is created](https://github.com/guzzle/guzzle/blob/master/src/Client.php#L331) (including for `Exception`s).

Did you happen to test this scenario?
This would explain a super-annoying thing I had with Laravel the other week, when the `onFulfilled` callback received an `Exception` instead of the `Response` I was expecting.

Sammyjo20:
Yes, so it follows the A+ standard for promises, one of the rules states that if something is thrown in the reject handler, it will be passed to the reject handler of the next "then", however if something is **returned** in the reject handler, it will be passed to the resolve of the next handler, how confusing! So sadly I can only have an exception in the "reject"...

juse-less:
Thanks for the clarification. That does explain the issue I had.
I can see the use for either case. But both works for me.

Thanks for looking into async requests so quickly. 🙂

Sammyjo20:
I was going to see if I can also roll pools into this too, as it would help the other guy who originally posted the issue. 

No problem! I wanted to tackle it but I didn’t know what it was admittedly, so thank you for the learning opportunity!

Sammyjo20 (commented):
Self review complete

PR #59, opened 2022-05-22
Feature | Nested SDK Requests

Sammyjo20:
Previously you could define custom methods on the connector to call requests in Saloon.

```php
protected array $requests = [
    'getMyUser' => UserRequest::class,
    ErrorRequest::class,
];
```

This was pretty limited as you couldn't define "groups" of requests. 

This PR introduces an extension to this API and now you can define an array of requests, keyed for a collection or a custom "RequestCollection" which is a class that extends the base "RequestCollection" class and can have completely custom methods inside. You will also have access to the connector.

### Define groups of requests

`$connector->user()->get()`

```php
protected array $requests = [
    'user' => [
        'get' => UserRequest::class,
    ],
];
```

### Define custom request collections

`$connector->user()->get()`

```php
protected array $requests = [
    'user' => UserCollection::class,
];
```

```php
class UserCollection extends RequestCollection
{
    public function get(): SaloonRequest
    {
        return $this->connector->request(new UserRequest);
    }
}
```

### DIFF
--- a/src/Exceptions/InvalidRequestKeyException.php
+++ b/src/Exceptions/InvalidRequestKeyException.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Sammyjo20\Saloon\Exceptions;
+
+class InvalidRequestKeyException extends SaloonException
+{
+    //
+}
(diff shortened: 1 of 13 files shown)


Sammyjo20 (commented):
Self review complete

Gummibeer:
`'user' => UserCollection::class,` this is neat! 🚀

PR #63, opened 2022-06-10
Feature | Spring cleaning & added authentication alias

Sammyjo20:
This PR is just some nice spring cleaning, making sure all requests/connectors return "static" and not "self" to improve developer experience, and also added an alias so now you can write:

```php
$request = new GetForgeServersRequest;
$request->authenticate($auth);
```


Sammyjo20 (commented):
Self review complete

PR #67, opened 2022-06-13
feature/closure mock responses

Sammyjo20:
- Added closure mocking
- Added test for callable mocking


Sammyjo20 (commented):
Self review complete

PR #70, opened 2022-06-22
Make PluginManager respect plugin traits used by request & connector's parent class

leo108:
Says there is a base request:

```
class BaseRequest extends SaloonRequest {
    use HasJsonBody;
}
```

And a subclass:

```
class SubRequest extends BaseRequest {
    public function defaultData(): array {
        return ['foo' => 'bar'];
    }
}
```

```
$requestManager = new RequestManager(new SubRequest());
$requestManager->hydrate();
// Before: null
// After: ["foo" => "bar"]
dump($request->getConfig('json'));

```

We can enhance this quite easily thanks to Laravel's helper functions.

### DIFF
--- a/src/Traits/ManagesPlugins.php
+++ b/src/Traits/ManagesPlugins.php
@@ -2,8 +2,6 @@
 
 namespace Sammyjo20\Saloon\Traits;
 
-use ReflectionClass;
-
 trait ManagesPlugins
 {
     /**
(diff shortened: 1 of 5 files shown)


Sammyjo20 (approved):

Sammyjo20:
This works really well, thanks for the contribution, @leo108 !

PR #71, opened 2022-07-11
Feature | Fixture Recorder

Sammyjo20:
This PR introduces a "fixture recorder" which allows you to record your requests in your tests the first time they are made, and then it will store the fixture for later so it won't make the real request each time.

You will be able to do this in your tests:

```php
$recorder = new FixtureRecorder($path);
$response = $recorder->record($request, $connector, $customName);

// In your Pest.php

function recorder()
{
    return new FixtureRecorder($fixturePath);
}

// In your tests... Will be automatically replaced!

$response = recorder()->record(new MyRequest);
```

See here too: https://www.juststeveking.uk/testing-api-responses-in-pest-php/

Going to write up the docs and then will merge this in!

Todo:
- [ ] Introduce a new method `recordFailure()` to make it simple and remove the failure option on a per-instance basis
- [ ] FOR LARAVEL: Recorder ready in the container so it will record any requests that you specify
- [ ] Have a good think about how it would work. Does it replace the current MockClient? Could you add a recorder to the existing mock client?

Alternative recorder syntax? Only accept request instances, so if you want to record a connector making a request you do

$recorder->record($connector->request($request));

Other things to consider:
- What if I want to record requests happening in nested code 
- Maybe build it on top of the mock client 

Idea for Laravel:

```php
Saloon::fixtureRecorder([
    new RecordedResponse(Request::class, 'custom-name', 'allow-failures')
]);

// Or

Saloon::fixtureRecorder()
   ->recordRequest(RequestA::class, 'custom-name', 'allow-failures')
   ->recordRequest(RequestB::class, 'custom-name', 'allow-failures')

// Laravel will then store those custom recorded responses if it has been defined.
```


Sammyjo20 (commented):
Self review complete

PR #76, opened 2022-07-22
Feature | Added new resolveConnector method

Sammyjo20:
Fixes https://github.com/Sammyjo20/saloon-laravel/issues/22

### DIFF
--- a/src/Http/SaloonRequest.php
+++ b/src/Http/SaloonRequest.php
@@ -107,7 +107,18 @@ abstract class SaloonRequest implements SaloonRequestInterface
             throw new SaloonInvalidConnectorException;
         }
 
-        $this->setConnector(new $this->connector);
+        $this->setConnector($this->resolveConnector($this->connector));
+    }
+
+    /**
+     * Resolve the connector.
+     *
+     * @param string $connectorClass
+     * @return SaloonConnector
+     */
+    protected function resolveConnector(string $connectorClass): SaloonConnector
+    {
+        return new $connectorClass;
     }
 
     /**


rrolla (approved):
Looks nice :)

juse-less:
Looks good to me as well. 👍 

Just a small note: The property is nullable (and null by default), so passing it as an argument without the developer also setting the property, would end in an Exception.

For v2, if possible, it'd be nice if you could figure out a way to only need to provide one or the other.
Currently we need to set both the property, and resolve the class in it as well, which is a bit of a duplication, albeit not a big one.
Without breaking backwards compatibility in v1, that's a bit trickier, though.

Sammyjo20:
Good shout @juse-less, I think it should be okay since in the `bootConnector` method I validate the connector property. I was thinking about what you said and I could easily implement a pattern similar to how Laravel Livewire's $rules property works.

You will be able to define a $connector property, or you could provide a `defineConnector` method where you can resolve any class, but it has to be an instance of SaloonConnector.

juse-less:
> I think it should be okay since in the bootConnector method I validate the connector property.

You are absolutely right. I didn't reflect on that. 👍
Looks good to me. 🙁

juse-less:
Well.. this is a bit awkward, given I approved it.

I'm in the midst of updating all HTTP code, and realised that this wasn't fully working (for me), after all.
As I'm binding interfaces to classes, `class_exists()` will (does) cause [the check](https://github.com/Sammyjo20/Saloon/blob/b0387889b10192e8e58e6dcf26325ce138a415c8/src/Http/SaloonRequest.php#L100-L102) to fail.
I quickly tested in Tinkerwell, as I wasn't quite sure why it was failing, and quickly realised that `class_exists()` will literally check for classes, not interfaces, which `interface_exists()` does.
Could've sworn `class_exists()` checks all 'classy' things, but had forgotten about `interface_exists()` until I just tested.

Even if `interface_exists()` is also checked, [there's a check](https://github.com/Sammyjo20/Saloon/blob/b0387889b10192e8e58e6dcf26325ce138a415c8/src/Http/SaloonRequest.php#L104) if the `SaloonRequest::$connector` is an instance of `SaloonConnector`, before the interface has been resolved in the service container.

@Sammyjo20 Want me to submit this as a new feature request, or so?
It's not urgent in any means, but thought I'd write it down here now that I noticed it.

Issue #86, opened 2022-10-07
[v2] Drop illuminate/support requirement


Gummibeer:
As already discussed on Twitter - the `illuminate/support` package comes with some handy classes but also a ton of Laravel-specific ones. Requiring it only because of the `Arr` and/or `Str` helpers isn't really "right" and questions the "plain php" nature of that package, as there's a specific Laravel wrapper.

It will also tie this package's versioning to the Laravel lifecycle even if there hasn't been anything changed relevant to that package. As otherwise it won't be installable on the latest Laravel projects - at the same time it should support the lowest Illuminate version possible as with Illuminate and Laravel comes also PHP constraints which will possibly limit the usability of the package even further. In my experience Symfony devs also don't really like to load illuminate/support as it comes with a ton of classes absolutely not needed and will make navigating IDE more complicated.

So far I've seen the only classes used are the following. All should be "easy" to replace with a custom `Util` class or even plain PHP.

* `Illuminate\Support\Arr`
* `Illuminate\Support\Str`
* `Illuminate\Support\Collection`

https://github.com/Sammyjo20/Saloon/search?q=Illuminate

Wulfheart:
Aren’t there also packages for each of this? So we could just include Collection?

Gummibeer:
`Collection` is the only standalone one and is also only used for the `collect()` method. This could even be solved by a Composer `suggest` thing as it's a standalone and encapsulated method not used anywhere in the package but only a public shorthand.
https://github.com/Sammyjo20/Saloon/blob/e3f6e19daba415dd4907ada3164b77a44d3ff8f1/src/Http/SaloonResponse.php#L182-L191

And once in the `ProxyRequestNameHelper` which can be easily rewritten to a `foreach()` loop.
https://github.com/Sammyjo20/Saloon/blob/8efb063a04bd2263d26d50790ebae801bb864d22/src/Helpers/ProxyRequestNameHelper.php#L21-L37

Jimlah:
@Gummibeer can you assign it to me. I will love to work on it.

Gummibeer:
@Sammyjo20 will have to assign it - but I won't work on it in that cas. So go ahead! :)

Sammyjo20:
Will be adding these changes to v2 👍

PR #87, opened 2022-10-07
Drop illuminate/support requirement

Jimlah:
#86

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^3.5",
-        "illuminate/support": "^9.34",
+        "illuminate/collections": "^9.0",
         "pestphp/pest": "^1.21",
         "spatie/ray": "^1.33"
     },
@@ -42,7 +42,7 @@
         }
     },
     "suggest": {
-        "illuminate/support": "Needs to support collect method in class Saloon Response"
+        "illuminate/collections": "Needs to support collect method in class SaloonResponse"
     },
     "config": {
         "allow-plugins": {
--- a/composer.json
+++ b/composer.json
@@ -21,11 +21,12 @@
     "require": {
         "php": "^8.0",
         "guzzlehttp/guzzle": "^7.4",
-        "illuminate/support": "^8.0 || ^9.0",
-        "nesbot/carbon": "^2.58"
+        "nesbot/carbon": "^2.58",
+        "ext-simplexml": "*"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^3.5",
+        "illuminate/support": "^9.34",
         "pestphp/pest": "^1.21",
         "spatie/ray": "^1.33"
     },
@@ -39,6 +40,9 @@
         "psr-4": {
             "Sammyjo20\\Saloon\\Tests\\": "tests/"
         }
+    },
+    "suggest": {
+        "illuminate/support": "Needs to support collect method in class Saloon Response"
     },
     "config": {
         "allow-plugins": {
--- a/src/Helpers/StateHelper.php
+++ b/src/Helpers/StateHelper.php
@@ -2,17 +2,27 @@
 
 namespace Sammyjo20\Saloon\Helpers;
 
-use Illuminate\Support\Str;
-
 class StateHelper
 {
     /**
      * Generate a random string for the state.
      *
      * @return string
+     * @throws \Exception
      */
     public static function createRandomState(): string
     {
-        return Str::random(32);
+        $string = '';
+        $length = 32;
+
+        while (($len = strlen($string)) < $length) {
+            $size = $length - $len;
+
+            $bytes = random_bytes($size);
+
+            $string .= substr(str_replace(['/', '+', '='], '', base64_encode($bytes)), 0, $size);
+        }
+
+        return $string;
     }
 }
--- a/src/Helpers/URLHelper.php
+++ b/src/Helpers/URLHelper.php
@@ -2,8 +2,6 @@
 
 namespace Sammyjo20\Saloon\Helpers;
 
-use Illuminate\Support\Str;
-
 class URLHelper
 {
     /**
@@ -15,7 +13,7 @@ class URLHelper
      */
     public static function matches(string $pattern, string $value): bool
     {
-        return Str::is(Str::start($pattern, '*'), $value);
+        return self::is(self::start($pattern, '*'), $value);
     }
 
     /**
@@ -37,5 +35,42 @@ class URLHelper
         $baseEndpoint = $requiresTrailingSlash ? $baseEndpoint . '/' : $baseEndpoint;
 
         return $baseEndpoint . $endpoint;
+    }
+
+    /**
+     * Check if the url pattern matches the value
+     *
+     * @param string $pattern
+     * @param string $value
+     * @return bool
+     */
+    public static function is(string $pattern, string $value): bool
+    {
+        if ($pattern === $value) {
+            return true;
+        }
+
+        $pattern = preg_quote($pattern, '#');
+        $pattern = str_replace('\*', '.*', $pattern);
+
+        if (preg_match('#^' . $pattern . '\z#u', $value) === 1) {
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
+     * Give a url a prefix
+     *
+     * @param string $value
+     * @param string $prefix
+     * @return string
+     */
+    public static function start(string $value, string $prefix): string
+    {
+        $quoted = preg_quote($prefix, '/');
+
+        return $prefix.preg_replace('/^(?:'.$quoted.')+/u', '', $value);
     }
 }
--- a/src/Http/SaloonResponse.php
+++ b/src/Http/SaloonResponse.php
@@ -2,12 +2,12 @@
 
 namespace Sammyjo20\Saloon\Http;
 
+use Exception;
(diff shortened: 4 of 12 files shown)


Jimlah:
@Gummibeer Here is the PR for the #86

Sammyjo20:
Thanks @Jimlah I will review this weekend. Don't worry if the code style fails.

composer.json
```diff
@@ -21,11 +21,12 @@
...
         "php": "^8.0",
         "guzzlehttp/guzzle": "^7.4",
-        "illuminate/support": "^8.0 || ^9.0",
-        "nesbot/carbon": "^2.58"
+        "nesbot/carbon": "^2.58",
+        "ext-simplexml": "*"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^3.5",
+        "illuminate/support": "^9.34",
```
Gummibeer: as `9.34` isn't really required but `^9.0` works as well I would go for that constraint.
PS: if it's only for the `collect()` method it would be [`illuminate/collections`](https://github.com/illuminate/collections) instead of the whole `illuminate/support` package.

Jimlah: I will work on it and send a new PR

Gummibeer: Don't send a new PR - just commit to your already existing branch. It will update this PR. 😉

Jimlah: Done

tests/Feature/Oauth2/AuthCodeFlowConnectorTest.php
```diff
@@ -51,7 +50,7 @@
 
     expect($state)->toBeString();
 
-    expect(Str::endsWith($url, $state))->toBeTrue();
+    expect(str_ends_with($url, $state))->toBeTrue();
```
Sammyjo20, line 53: I can't see where this method has been defined, is this a typo? Or is this a built-in pest function?

src/Http/SaloonResponse.php
```diff
@@ -182,12 +185,18 @@ public function xml(...$arguments): SimpleXMLElement|bool
...
+     * @return mixed
+     * @throws Exception
      */
-    public function collect($key = null): Collection
+    public function collect($key = null): mixed
     {
-        return Collection::make($this->json($key));
+        $collection = 'Illuminate\\Support\\Collection';
+        if (! class_exists($collection)) {
+            throw new ClassNotFoundException($collection);
```
Sammyjo20, line 196: Would we also be able to add a custom message to this to suggest installing `illuminate/collection` please? We may need our own exception for this. What do you think?

src/Http/SaloonResponse.php
```diff
@@ -150,8 +150,11 @@ public function json($key = null, $default = null)
         if (is_null($key)) {
             return $this->decodedJson;
         }
+        if (array_key_exists($key, $this->decodedJson)) {
+            return $this->decodedJson[$key];
+        }
 
-        return Arr::get($this->decodedJson, $key, $default);
+        return [];
```
Sammyjo20, line 157: Ah, I actually think we're losing a bit of functionality if we switch back to using normal array access. With `Arr::get` you can do cool things like `nestedObject.value` and it will return the value from the dot-notion. Is it possible if we could extract Arr::get into our own helper to keep this logic?

src/Traits/CollectsConfig.php
```diff
@@ -87,8 +85,8 @@ public function getConfig(string $key = null): mixed
             ? array_merge($this->defaultConfig(), $this->customConfig)
             : $this->customConfig;
 
-        if (isset($key)) {
-            return Arr::get($configBag, $key);
+        if (isset($key) && array_key_exists($key, $configBag)) {
+            return $configBag[$key];
```
Sammyjo20, line 89: Same comment regarding Arr::get on src/Http/SaloonResponse.php

src/Traits/CollectsData.php
```diff
@@ -101,8 +100,8 @@ public function getData(string $key = null): mixed
             $dataBag = $this->customData;
         }
 
-        if (isset($key)) {
-            return Arr::get($dataBag, $key);
+        if (isset($key) && array_key_exists($key, $dataBag)) {
+            return $dataBag[$key];
```
Sammyjo20, line 104: Same comment regarding Arr::get on src/Http/SaloonResponse.php

src/Traits/CollectsHeaders.php
```diff
@@ -86,8 +84,8 @@ public function getHeaders(string $key = null): mixed
             ? array_merge($this->defaultHeaders(), $this->customHeaders)
             : $this->customHeaders;
 
-        if (isset($key)) {
-            return Arr::get($headerBag, $key);
+        if (isset($key) && array_key_exists($key, $headerBag)) {
+            return $headerBag[$key];
```
Sammyjo20, line 88: Same comment regarding Arr::get on src/Http/SaloonResponse.php

src/Traits/CollectsQueryParams.php
```diff
@@ -97,8 +96,8 @@ public function getQuery(string $key = null): mixed
             $queryBag = $this->customQuery;
         }
 
-        if (isset($key)) {
-            return Arr::get($queryBag, $key);
+        if (isset($key) && array_key_exists($key, $queryBag)) {
+            return $queryBag[$key];
```
Sammyjo20, line 100: Same comment regarding Arr::get on src/Http/SaloonResponse.php

src/Helpers/URLHelper.php
```diff
@@ -38,4 +36,41 @@ public static function join(string $baseUrl, string $endpoint): string
...
+        }
+
+        $pattern = preg_quote($pattern, '#');
+        $pattern = str_replace('\*', '.*', $pattern);
+
+        if (preg_match('#^' . $pattern . '\z#u', $value) === 1) {
+            return true;
+        }
+
+        return false;
```
Sammyjo20, line 60: How come we're not using the same code as Illuminate/support Str here? https://github.com/illuminate/support/blob/master/Str.php#L357

src/Helpers/StateHelper.php
```diff
@@ -2,17 +2,27 @@
...
+
+        while (($len = strlen($string)) < $length) {
+            $size = $length - $len;
+
+            $bytes = random_bytes($size);
+
+            $string .= substr(str_replace(['/', '+', '='], '', base64_encode($bytes)), 0, $size);
+        }
+
+        return $string;
```
Sammyjo20, line 26: Same here, I kind of wish we had the same functionality as: https://github.com/illuminate/support/blob/master/Str.php#L726

Sammyjo20 (changes requested):
Hey, thanks for your PR - it's much appreciated.

Apologies for lots of comments, and great work but I'm having a few thoughts and I'm not sure if I necessarily want to remove Illuminate/Support from Saloon. There are so many helper methods that are useful and while it's great to have alternatives inside Saloon I feel it's extra code to continuously maintain when Laravel and the core team will likely continue maintaining illuminate/support for a very long time. 

I kind of wish there were smaller helpers where we could bring in just "String helpers" or just "collections" but I don't think this is a thing.

So in summary, great PR but I think it's worth discussing if it's worth introducing this much code into the core just to remove a dependant?

Sammyjo20:
@Gummibeer Tagging you for your thoughts too.

Sammyjo20:
I've re-read the issue that you posted on #86 and I understand a bit more now why it's useful to remove it.

Jimlah:
While working on it. I did wished we could have left the illuminate support, because of the many helper function it supported. But then again I could see that along that the helper function where not fully used. And about the collection, most non laravel users may not fully know about it. Untill they read about it.

tests/Feature/Oauth2/AuthCodeFlowConnectorTest.php
Gummibeer: It's a native PHP function. 
https://www.php.net/manual/de/function.str-ends-with.php

Gummibeer:
Regarding the reason to remove it: the support package is one of the fundamental base packages of Laravel framework. Even the support package itself has a bunch of dependencies - following as of writing:
```
        https://github.com/doctrine/inflector: "^2.0",
        https://github.com/illuminate/collections: "^10.0",
        https://github.com/illuminate/conditionable: "^10.0",
        https://github.com/illuminate/contracts: "^10.0",
        https://github.com/illuminate/macroable: "^10.0",
        https://github.com/briannesbitt/Carbon: "^2.53.1",
        https://github.com/voku/portable-ascii: "^2.0"
```

These alone will quickly sum up to a pretty big tree of dependencies that can conflict with the package itself and/or the plain PHP users of the package. Laravel also has a pretty modern PHP requirement all the time. So I know it's the ideal world but the reality is that a lot of PHP projects, primarily the plain PHP ones, stay on PHP7.4 for decades.

But also for Laravel users: you will have to upgrade it at least once per year and decide if you only add the new version or drop the old one. Adding will make it easier for users and is likely what's needed as the used helpers don't really change since years. But this will increase your maintenance work. As you have to support the whole tree of dependencies for every Laravel release. And this will possibly break your neck. As soon as a new PHPUnit major version or PHP major version is released and used by Laravel you will have a painful time.

I've been there with @astrotomic / laravel-translatable to support as many PHP/Laravel versions as possible. It's absolutely painful! And for a plain PHP package having this trouble is absolutely not needed.

Sammyjo20:
Thank you for your effort @Jimlah however I've decided that I'm going to hold off on these changes until v2, since it requires breaking changes to remove the Carbon package and Illuminate Support package, e.g for the `collect` method.

PR #90, opened 2022-10-23
Feature | Mocking Fixtures

Sammyjo20:
This PR introduces a new feature into Saloon called "mock fixtures"! This feature is intended for people who want to only make a real request to an API once and then store/cache the response for all future tests. This is now easy with the new fixture mock response.

## Example

```php
<?php

$mockClient = new MockClient([
    // You specify a name of the fixture, e.g "singleUser". You may also namespace the fixture like `socialnetwork/singleUser`
    UserRequest::class => MockResponse::fixture('singleUser'),
]);

// This will make the first API call and record the fixture.
// If it doesn't exist, the response from this API call will
// be stored in tests/Fixtures/Saloon.

$responseA = UserRequest::make()->send($mockClient);

// Now if the request is called again, and a fixture exists,
// the request will not be sent again but the fixture 
// response will be used instead.

$responseB = UserRequest::make()->send($mockClient);
```

## How does it work?

When you use `MockResponse::fixture()`, Saloon will check if a fixture file exists in the fixture directory, the file is a JSON file containing the status code, headers and body. If the file exists then the file will be retrieved and a MockResponse will be automatically created. If the file doesn't exist, it will let the real request go through and then record the file for future use.

## What options are there?

You can configure the fixture directory as well as if the mock client should throw an exception if a fixture doesn't exist.

`MockConfig::setFixturePath()`
`MockConfig::throwOnMissingFixtures()`


Sammyjo20 (commented):
Self review

Gummibeer:
How about a fixture fallback name like done by spatie in the snapshot assertion? 🤔
https://github.com/spatie/phpunit-snapshot-assertions/blob/0141b5cfc18da18c76081722ca5f055672e5b9d0/src/MatchesSnapshots.php#L161-L165
```php
protected function getSnapshotId(): string
{
    return (new ReflectionClass($this))->getShortName().'__'.
        $this->getName().'__'.
        $this->snapshotIncrementor;
}

protected function getSnapshotDirectory(): string
{
    return dirname((new ReflectionClass($this))->getFileName()).
        DIRECTORY_SEPARATOR.
        '__snapshots__';
}
```

src/Http/Fixture.php
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /**
+     * Attempt to get the mock response from the fixture.
+     *
+     * @return MockResponse|null
+     * @throws FixtureMissingException
+     * @throws \JsonException
+     */
+    public function getMockResponse(): ?MockResponse
```
Gummibeer, line 54: How about making that method request aware!? This would mean that the request could be passed to `getFixturePath()` and would lead to the option that `$this->name` could be a `Closure` that returns the fixture name `string` based on the current request being handled.

With that in place everyone would be totally free how to name the fixtures.

Gummibeer (changes requested):

src/Http/Fixture.php
Sammyjo20: Thanks for the review @Gummibeer 

My only concern with guessing the fixture names is that it could record unexpected results. For example, let's say I wrote a fixture like this:

```php
Saloon::fake([
    UserRequest => MockResponse::fixture(),
]);
```

That would be fairly easy to guess because the fixture name could just be the class name or the class's file path, similar to what Spatie does in their snapshot assertions package. 

My concern is when something like this happens

```php
Saloon::fake([
    MockResponse::fixture(),
]);

// or

Saloon::fake([
    '*' => MockResponse::fixture(),
]);
```

Since the fixture doesn't have name, if we were to make it request aware, it would pick up the very first request and store the fixture data against that request without the user really knowing that it was recording that request. If someone then passes a different request in, it will record that one separately as well or if they happened to have the same request, it would return that request's fixture.

I think the issue is that from a readability perspective just `MockResponse::fixture()` doesn't say anything about the request, however if it's something like this

```php
Saloon::fake([
    MockResponse::fixture('userMock'),
]);
```

It becomes similar to PHP-VCR's "cassette" functionality where the user explicitly requests a given mock. 

By the way, you can still use a closure if you would like to auto-generate the name for you, for example

```php
Saloon::fake([
    '*' => function (SaloonRequest $request) {
           return MockResponse::fixture($request::class); // Provide any name here
    }
]);
```

Sammyjo20: I have provided a better example where you can define your own name generator for fixtures. I think this is what you are after? 🤠

```php
<?php

$mockClient = new MockClient([
    '*' => function (SaloonRequest $request) {
        $reflection = new ReflectionClass($request);

        return MockResponse::fixture($reflection->getShortName());
    },
]);

// This will create a fixture called "GetForgeServerRequest"

$request = new GetForgeServerRequest(12345);
$response = $request->send($mockClient);

// This will create a fixture called "GetAllForgeServersRequest"

$request = new GetAllForgeServersRequest($data);
$response = $request->send($mockClient);
```

Gummibeer: Regarding guessing and spatie: if you check the link on depth:
It uses the test class and method name plus an Index. So every request execution will have its own name based on the test it was executed in.

Regarding the anonymous function: yes, that's exactly it. Just had it switched in order in mind.

```php
Mock response::fixture(fn(Request $request) => ...)
```

But result is the same so totally fine for me! 🥳

Gummibeer:
> Failed asserting that exception of type "Sammyjo20\Saloon\Exceptions\DirectoryNotFoundException" matches expected exception "App\Clients\Avalara\Exceptions\Client\ModelStateInvalidException". Message was: "The directory "tests/Fixtures/Saloon" does not exist or is not a valid directory." 

I get the following error - seems like I have to create that folder manually. Couldn'T you create it with `@mkdir($dir, 0777, true)`? 🤔

PR #91, opened 2022-10-25
Feature | Reduce Saloon's Dependencies

### LINKED ISSUE #86: [v2] Drop illuminate/support requirement

Sammyjo20:
Hey y'all, I wanted to write a follow-up PR regarding issue #86 and the PR #87 

This PR still needs tests for the helper functions but I've essentially removed the `illuminate/support` requirement in favour of extracting the methods Saloon uses, this significantly reduces Saloon's dependency footprint. I have also removed the Carbon package since it was only being used for the OAuth2 authentication.

I have suggested `illuminate/collections` and `nesbot/carbon` since existing code won't work without those packages. 

I still had to introduce `spatie/macroable` as the SaloonResponse uses the Macroable trait, and it's a dependency-free package so I am happy to keep that.

@Gummibeer @jimlah, what do you think?

My only question is around semantic versioning, do we have to wait for v2 to make these changes since it requires breaking changes (i.e you have to install specific packages) or is there somehow I can still release this without breaking changes?

Closes #86 
Closes #87


Sammyjo20 (commented):
Self review

Gummibeer:
Hey,
regarding BCs this is a nearly philosophical question. Everyone using the functionality in need of `illuminate/support`, `illuminate/collection` or `nesbot/carbon` would have them installed in their own dependencies.
As no one could instantiate a `Carbon` without having it as a dependency - relying on it being installed by another dependency is a very bad practice as dependencies aren't part of the breaking contract.

So officially: no one could have breaking problems and in case they would have problems with their own dependency management.

Second: the public API remains the same. So everything can be called the same way as before.

So for me, even if there's a 1% chance for one project having breaking problems - I, as a maintainer, would just blame them and ignore it. But you are the maintainer so it's totally up to you.

Sammyjo20:
Hey, that's a good point - I'm not so worried about people using Carbon because Saloon installed it, I'm more worried about them not having Carbon/Collections installed and their code uses the `$response->collect()` method, or it uses the OAuth2 logic which uses Carbon.

What are your thoughts around that? Because as a consumer of the package, if I updated it and then suddenly `collect` stopped working, I would blame the package maintainers for making a breaking change.

Gummibeer:
If we are talking about this: shouldn't the `collect()` method be part of `saloon-laravel`!? 🤔

In that case you would go fully breaking and move it whefe it belongs to.

Regarding carbon: haven't checked the code but am pretty sure that you could achieve the same with `DateTimeInterface` as I assume the plain PHP `fornat()` and `modify()` methods can solve all you need.

Sammyjo20:
I think the best thing to do is wait until v2 to remove these dependencies, so I will close the PR for now.

PR #92, opened 2022-10-25
Fix | Leading Slash When Base URL is empty and endpoint in request is a URL

### LINKED ISSUE #26: BaseUrl gets an '/' at the beginning if empty
Hello guys,

thank you for this wonderful package.

I am developing a connector for Google OAuth.

Google provides three completely different Base URL's for Authorization, Token and User.

My plan was to define each complete URL as an Endpoint in the OAuthConfig and leave the BaseUrl empty.

Doing so results in a request URL that always starts with an '/' in the beginning which makes the curl fail.

For example:

`$connector->getAuthorizationUrl() // =>   /https://accounts.google.com/o/...`

Same is true for The Token & User Endpoint.

Now I know I could create 3 different connectors for each endpoint and provide a non empty BaseURL and the Endpoint but then I would have 3 files instead of one and it kinda fails the purpose of this flow in my opinion.

I think it's an easy fix to skip the beginning '/' if the BaseUrl is empty.

```
class YoutubeOAuthConnector extends SaloonConnector
{
    use AuthorizationCodeGrant;

    public function defineBaseUrl(): string
    {
        return '';
    }

    protected function defaultOauthConfig(): OAuthConfig
    {
        return OAuthConfig::make()
            ->setClientId(config('youtube.api.client_id'))
            ->setClientSecret(config('youtube.api.client_secret'))
            ->setDefaultScopes(config('youtube.api.scopes'))
            ->setAuthorizeEndpoint('https://accounts.google.com/o/oauth2/v2/auth?access_type=offline&prompt=consent')
            ->setTokenEndpoint('https://oauth2.googleapis.com/token')
            ->setUserEndpoint('https://www.googleapis.com/oauth2/v3/userinfo')
            ->setRedirectUri(route(config('youtube.api.redirect_uri')));
    }
}
```

Sammyjo20:
This PR fixes https://github.com/Sammyjo20/saloon-laravel/issues/26

### DIFF
--- a/src/Helpers/URLHelper.php
+++ b/src/Helpers/URLHelper.php
@@ -34,7 +34,11 @@ class URLHelper
         $requiresTrailingSlash = ! empty($endpoint) && $endpoint !== '/';
 
         $baseEndpoint = rtrim($baseUrl, '/ ');
-        $baseEndpoint = $requiresTrailingSlash ? $baseEndpoint . '/' : $baseEndpoint;
+
+        $endpointIsNotUrl = empty(filter_var($endpoint, FILTER_VALIDATE_URL));
+        $glue = $endpointIsNotUrl ? '/' : null;
+
+        $baseEndpoint = $requiresTrailingSlash ? $baseEndpoint . $glue : $baseEndpoint;
 
         return $baseEndpoint . $endpoint;
     }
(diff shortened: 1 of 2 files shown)


Sammyjo20 (commented):
Self review

PR #97, opened 2022-11-03
automatically create fixture directory

Gummibeer:
This PR automatically creates the fixture base directory if it does not exist. As before it was impossible to fix the directory and use it in the same runtime.
And directories aren't anything irreversible like foreign API calls (😉).


Gummibeer:
Missed far too many spots to change.

Issue #99, opened 2022-11-07
`URLHelper::join()` does not respect full-url in `$endpoint`


Gummibeer:
The `\Sammyjo20\Saloon\Helpers\URLHelper::join()` method does not respect a full URL in the `$endpoint` argument but still prefixes it with the `$baseUrl`.
Some APIs have a few endpoints on another base URL/domain. I would like to override the default base URL with a full URL in the request endpoint instead of creating a whole new connector.

The already used `filter_var($endpoint, FILTER_VALIDATE_URL)` could be used to early return in case the `$endpoint` is already a full and valid URL. Possibly also combined with `parse_url()` to check scheme and host are defined.

Right now I'm running in the following curl error which fully bubbles through all layers. Even with the `AlwaysThrowsOnErrors` trait it fully bubbles through and doesn't get wrapped up in a `\GuzzleHttp\Exception\ConnectException` or similar.

> cURL error 6: Could not resolve host: `api.steampowered.comhttps` (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for `https://api.steampowered.comhttps//steamcommunity.com/actions/QueryLocations?key=xyz&format=json`

```php
class SteamConnector extends SaloonConnector
{
    public function defineBaseUrl(): string
    {
        return 'https://api.steampowered.com';
    }
}
```

```php
class QueryLocationsRequest extends SaloonRequest
{
    protected ?string $method = 'GET';

    public function defineEndpoint(): string
    {
        $query = '';

        if ($this->countryCode) {
            $query .= "/{$this->countryCode}";

            if ($this->stateCode) {
                $query .= "/{$this->stateCode}";
            }
        }

        return "https://steamcommunity.com/actions/QueryLocations{$query}";
    }
}
```

ashleyhood:
I agree that if the `\Sammyjo20\Saloon\Helpers\URLHelper::join()` method could check if the `$endpoint` is a correctly formed URL and if so, exit early and return the `$endpoint` as the URL.

This would help me, as at the moment when connecting to the [Xero Accounting API](https://developer.xero.com/documentation/guides/oauth2/auth-flow/) I have to create two connectors. One connector to authorize and another connector to get the token because Xero has different base URLs for each:

- Authorize: `https://login.xero.com`
- Token: `https://identity.xero.com`

This means I would just be able to set the `OAuthConfig::make()->setTokenEndpoint()` to the Xero API URL token endpoint without creating a separate connector.

[...]

PR #101, opened 2022-11-07
enhance: UrlHelper join method handle endpoint URL string

ashleyhood:
This is to allow API endpoints to be full URLs and not have the baseUrl prepended to it. Closes issue #99


Sammyjo20 (approved):
Awesome, thank you!

Gummibeer:
Thanks! 🎉🚀

Issue #64, opened 2022-06-11
Version 2 is in beta 🤠


Sammyjo20:
Hey everyone, thanks so much for all the support that you have given for Saloon. I can't believe it's almost at 500 stars on GitHub and receiving over 150 installs a day. I just wanted to take a moment to thank you all! 😀🙌

That being said, there are some good things coming to Saloon. I'm working on version 2, which will help create a road for the future of Saloon, as well as improving developer experience and making your life easier.

Here is a summary of the changes that are going to happen.

## New Flow

Currently, Saloon will run your request through the "Request Manager" which merges all of the headers, config, and everything else from the connector and request into one. With Version 2, I am introducing the "PendingSaloonRequest". Inside of here, this is where all of the logic like merging properties together and running your plugins will happen. This is so the building of a request is entirely separate from the sending of requests. After that, it will be sent to the "Guzzle Sender", and the class will receive the full PendingSaloonRequest with all the final configuration and headers before sending.

I am changing to this new flow because eventually, I want Saloon to be HTTP client agnostic and allow you to use any client you like, so you don't have to use Guzzle if you don't want - and if Guzzle decides to be abandoned, Saloon won't be left in the dark.

Here's the new flow in detail.

![image](https://user-images.githubusercontent.com/29132017/173206886-dc511ebc-257b-4614-8c79-110e3800aa71.png)

> The only exception to the flow in the image above is that it will not create a PSR-7 request just yet, I am still working that one out.

## Other Changes

- Brand new middleware pipeline so you don't have to rely on Guzzle's handler pipeline
- Requests don’t go straight to the request manager anymore, they can be built within a PendingSaloonRequest.
- Better way to send data, by implementing interfaces that are processed by the PendingSaloonRequest.
- All properties of a request like headers, config, data, handlers etc. Will all be standardised and now in separate buckets like `$request→headers()→push()`.
- Interfaces to be used a lot more
- PHP 8.1 minimum
- Real asynchronous request
- Request pooling support

## Middleware Pipeline

To help move the dependency on Guzzle, Saloon has also implanted its own middleware pipeline for requests and responses. This will replace the old Guzzle Handler support and response interceptors. You will be able to define request pipes and response pipes to modify the request/response before it is sent or given back to the user.

```php
$request = new GetForgeServersRequest;

$request->middleware()
    ->addRequestPipe(function (PendingSaloonRequest $request) {
       //
    })
    ->addRequestPipe(new MyInvokableClass)
[...]

PR #112, opened 2022-11-23
Fixed handler stack

Sammyjo20:
@Gummibeer I think you mentioned that Guzzle wasn't throwing exceptions on error requests, you were right - my old way of defining the handler stack meant the default middleware wasn't being applied. I've moved the logic to use the same as v1, and it works again now!

### DIFF
--- a/src/Contracts/Response.php
+++ b/src/Contracts/Response.php
@@ -242,9 +242,9 @@ interface Response
      * Get a header from the response.
      *
      * @param string $header
-     * @return string|null
+     * @return string|array|null
      */
-    public function header(string $header): ?string;
+    public function header(string $header): string|array|null;
 
     /**
      * Close the stream and any underlying resources.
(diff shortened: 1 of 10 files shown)


Sammyjo20 (commented):
Self review

Gummibeer:
Not sure if this is fixed now - my problem was that I had a request that returned a 401 response, I got the exception (from my sdk response) but the fixture file wasn't created. So the test would do a real request every single time and needs a correct access token. Which isn't the sense of fixtures and wasn't like that in v1.

Sammyjo20:
This is what I mentioned in issue #64 - I think we need to have a middleware priority or the ability to add a middleware to the top of the stack - like the fixture recording middleware, so that it can still record the response and then throw the exception.

Gummibeer:
Sounds good - like request middlewares in Laravel.

PR #113, opened 2022-11-25
fix/v2 guzzle handler stack

Sammyjo20:
- Fixed handler stack
- Adding more tests
- Fixing tests
- wip
- All tests pass
- PHP CS Fixer


Sammyjo20 (commented):
Self review

PR #114, opened 2022-11-26
Fix | AlwaysThrowOnErrors Priority

Sammyjo20:
This fixes a bug with the `AlwaysThrowOnErrors` trait and the fixture recorder. Due to the traits being loaded before the fixture recorder - the exception was being thrown. I've added a new `prepend` parameter which will add a middleware to the top of the pipeline.

@Gummibeer - hope you are well! This should fix the issues you were having :)


Sammyjo20 (commented):
Self review

PR #115, opened 2022-11-26
Feature | Rename queryParameters to query

Sammyjo20:
- Renamed queryParameters to query
- Code style
- More tests


Sammyjo20 (commented):
Self review

PR #119, opened 2022-11-28
Feature | V2 Data-Object Responses

Sammyjo20:
This PR introduces a new interface and trait combination that you can use on your data objects which will instruct Saloon to set the original response onto the DTO automatically. This is especially useful when building SDKs that just return DTOs. Previously if you received a DTO, there would be no way to access the original response which was frustrating for the developer. 

This PR also implements the ability to specify a DTO as a property instead of overwriting the entire method - if you create a DTO and implement the `FromSaloonResponse` interface, it will automatically be cast 🎉


Sammyjo20 (commented):
Self review

PR #120, opened 2022-11-29
WIP: try to add generics

Gummibeer:
I wasn't able to get it done or further. In my understanding that's the right way. But somehow PHPStorm doesn't fully understand. It works as soon as I'm adding some methods like the `createPendingRequest` one to the `GetAppListRequest` class itself with the more precise `@return \Saloon\Contracts\Response<\Saloon\Tests\Integration\GetAppListRequest>`. But that's not the goal and should work without. 🤔

* https://phpstan.org/blog/generics-in-php-using-phpdocs
* https://psalm.dev/docs/annotating_code/templated_annotations/

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -5,8 +5,14 @@
          colors="true"
 >
     <testsuites>
-        <testsuite name="Test Suite">
-            <directory suffix="Test.php">./tests</directory>
+        <testsuite name="Unit Test Suite">
+            <directory suffix="Test.php">./tests/Unit</directory>
+        </testsuite>
+        <testsuite name="Feature Test Suite">
+            <directory suffix="Test.php">./tests/Feature</directory>
+        </testsuite>
+        <testsuite name="Integration Test Suite">
+            <directory suffix="Test.php">./tests/Integration</directory>
         </testsuite>
     </testsuites>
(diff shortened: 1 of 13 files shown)


Gummibeer:
I'm not sure if @nunomaduro or someone else more experienced with generics has any idea!? 🤔
The final goal would be that the return of `\Saloon\Tests\Integration\SteamConnector::getAppList` knows that it returns a response from the used request and because of that the `\Saloon\Traits\Responses\HasResponseHelpers::dto` method inherits the return type of `createDtoFromResponse()` of the used request class.

juse-less:
@Gummibeer Have you managed to figure it out?

Some of the API has changed, which made it easier with some parts, when I took a stab at it last week.
I'm not sure if it helps with the DTO-related things, though, as I didn't get that far.

Sammyjo20:
One thing I'd like to improve with generics, and I have no idea if it's possible, is the use of custom responses

Especially when you do something like 

```php
$connector = new Connector;
$connector->send(new Request); // Generics for the "Request"'s custom response?
```

Sammyjo20:
I'm going to close this PR for now as @juse-less has been adding generics along the way, inspired by this PR. Thanks for your effort @Gummibeer!

PR #127, opened 2022-12-11
Feature | V2 - Replace method with enum

Sammyjo20:
This PR replaces the old string-based `method` on requests with an enum to be strongly typed.


Sammyjo20 (commented):
Self review

PR #130, opened 2022-12-13
Feature | V2 Safe Body

Sammyjo20:
I noticed an issue which could come up and annoy developers a bit which is the requirement to add both the `WithBody` interface and a body trait like `HasJsonBody` to make the body work.

In previous versions of Saloon, the `data` method was always available. From v2, the data method has been replaced with `body` but it's not included on the request/connector by default. It is only added when adding a trait. I also thought it would be good to require an interface so instead of doing fancy `method_exist` checks in the PendingRequest, I could quite easily just check if either the connector or request is an instanceof WithBody. This has some added benefits of knowing for certain that the `body` method had a specific return type.

The issue is that if someone adds a body trait without the interface, Saloon will send the request without the body and without error. So I decided to add a trait onto each body trait that will throw an exception if it detects either the connector or request doesn't have the WithBody trait. 

@Gummibeer @juse-less what are your thoughts on this, is this too "magical"? Should I let developers make their own mistakes? I just worry this is the kind of thing that a developer could fall down a bit rabbit hole trying to work out why their request isn't sending any data.

### DIFF
--- a/src/Exceptions/BodyException.php
+++ b/src/Exceptions/BodyException.php
@@ -1,8 +1,9 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Saloon\Exceptions;
 
 class BodyException extends SaloonException
 {
-
 }
--- a/src/Exceptions/BodyException.php
+++ b/src/Exceptions/BodyException.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Saloon\Exceptions;
+
+class BodyException extends SaloonException
+{
+
+}
--- a/src/Traits/Body/ChecksForWithBody.php
+++ b/src/Traits/Body/ChecksForWithBody.php
@@ -1,10 +1,12 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Saloon\Traits\Body;
 
+use Saloon\Http\PendingRequest;
 use Saloon\Contracts\Body\WithBody;
 use Saloon\Exceptions\BodyException;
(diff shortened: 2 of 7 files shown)


Sammyjo20 (commented):
Self review

juse-less:
I can understand the problem, but from a code perspective -
* an interface is the _definition_, the 'contract'.
* the _accompanying_ trait contains the default _implementation_, for that definition.

Regardless, the implementation _should_ have the definition for having a body.

I guess you could check `method_exists()`, but it feels wrong in my book.
For `body()`, it doesn't really matter, but.. in normal circumstances, you could ask yourself 'how do you know it'll take the correct arguments, and returns what you'd expect by using `method_exists()`?'

So I'd say stick with both - but can understand if it's somewhat inconvenient for others.
Especially depending on levels of expertise.

Sammyjo20:
Thanks for your thoughts, that's why I also made it an interface - because the traits are optional, but if the pending request checks if a request uses a contract, then you could use any implementation you want - you don't have to use one of the body traits like in v1

Sammyjo20:
With the contract, at least Saloon knows what to expect from the body method.

juse-less:
I think what you've done in this PR makes sense, though - PHP will throw a fit if the `body()` method isn't implemented, while using the interface.
Using one of the 'Body' traits will instead nudge the user so they can add the interface to their request implementation, to prevent the request from silently being sent without a body.

PR #132, opened 2022-12-15
Fix | v2 Better Response Error Methods

Sammyjo20:
This PR changes the way that the response `throw` and `toException` methods work. Previously, it used the `failed()` method but now it uses a new `shouldThrowRequestException`.


Sammyjo20 (commented):
Self review

Sammyjo20:
@Gummibeer FYI regarding my DM, this is what I settled for in the end.

PR #133, opened 2022-12-16
Feature | V2 Better Body Tests + More Features

Sammyjo20:
This PR adds a couple more methods to the `ArrayBodyRepository` to make it consistent with `ArrayStore`. It also adds tests for all the body traits.


Sammyjo20 (commented):
Self review

PR #134, opened 2022-12-19
Feature | V2 MultipartValue Object

Sammyjo20:
## Todo
- [ ] I might be able to bring back the ArrayRepository interface

Dealing with multipart files isn't fun in Guzzle. Previously you would have to define an array like this.

```php
'multipart' => [
      [
          'name'     => 'foo',
          'contents' => 'data',
          'headers'  => ['X-Baz' => 'bar']
      ],
      [
          'name'     => 'baz',
          'contents' => Psr7\Utils::tryFopen('/path/to/file', 'r')
      ],
      [
          'name'     => 'qux',
          'contents' => Psr7\Utils::tryFopen('/path/to/file', 'r'),
          'filename' => 'custom_filename.txt'
      ],
];
```

Saloon's `HasMultipartBody` trait will now allow you to build up multipart objects easily. 

```php
$request = new UserRequest;

$request->body()->add(name: 'avatar', contents: '...', filename: 'avatar.png', headers: []);

$avatar = $request->body()->get('avatar'); // MultipartValue
```


Sammyjo20 (commented):
Self review

Sammyjo20 (commented):
Self review

src/Data/MultipartValue.php
```diff
@@ -0,0 +1,46 @@
...
+     * @param string|null $filename
+     * @param array $headers
+     */
+    public function __construct(
+        public string  $name,
+        public mixed   $value,
+        public ?string $filename = null,
+        public array   $headers = []
+    ) {
+        if (! $this->value instanceof StreamInterface && ! is_resource($value) && ! is_string($value) && ! is_numeric($value)) {
```
juse-less, line 27: Just a small suggestion -
When I have plenty of checks, I prefer to split them and group them by context, and only invert at the beginning, with a group of 'or-checks' instead.
That way I don't have to put 'extra effort' to find all the inversions - I know that they have to be one or the other, as I group them.

```suggestion
        if (! ($this->value instanceof StreamInterface || is_resource($value) || is_string($value) || is_numeric($value))) {
```

src/Helpers/Arr.php
```diff
@@ -71,4 +71,26 @@ public static function get(array $array, string|int|null $key, mixed $default =
...
+        foreach ($items as $key => $value) {
+            $assoc = $callback($value, $key);
+
+            foreach ($assoc as $mapKey => $mapValue) {
+                $result[$mapKey] = $mapValue;
+            }
+        }
+
+        return $result;
+    }
```
juse-less, line 95: Another small suggestion so you don't have to nest loops -
I think you can 'dumb' this down a bit, as we're expecting the `$callback` to return an array.

```suggestion
    public static function mapWithKeys(array $items, callable $callback): array
    {
        $result = [];

        foreach ($items as $key => $value) {
            $result += $callback($value, $key);
        }

        return $result;
    }
```

A side note, though  - the above will keep the first value, whereas the current behaviour will overwrite - so maybe `array_merge()` instead of union is better.

src/Repositories/Body/MultipartBodyRepository.php
```diff
@@ -4,10 +4,159 @@
...
+        $value = [];
+
+        foreach ($arrays as $array) {
+            $value[] = $this->parseMultipartArray($array);
+        }
+
+        $this->data->merge(...$value);
+
+        return $this;
+    }
```
juse-less: It should be possible to dumb this down, to a more declarative approach.

```suggestion
    public function merge(array ...$arrays): static
    {
        $this->data->merge(...array_map(
            $this->parseMultipartArray(...),
            $arrays,
        ));

        return $this;
    }
```

src/Traits/Body/ChecksForWithBody.php
```diff
@@ -19,10 +19,10 @@ trait ChecksForWithBody
      */
     public function bootChecksForWithBody(PendingRequest $pendingRequest): void
     {
-        if ($pendingRequest->getRequest() instanceof WithBody || $pendingRequest->getConnector() instanceof WithBody) {
+        if ($pendingRequest->getRequest() instanceof HasBody || $pendingRequest->getConnector() instanceof HasBody) {
             return;
         }
 
-        throw new BodyException(sprintf('You have added a body trait without adding the `%s` interface to your request/connector.', WithBody::class));
+        throw new BodyException(sprintf('You have added a body trait without adding the `%s` interface to your request/connector.', HasBody::class));
```
juse-less: I _think_ request/connector could get confused for the need to exist on both the request and connector.

```suggestion
        throw new BodyException(sprintf('You have added a body trait, without adding the `%s` interface to your request or connector.', HasBody::class));
```

I'm also a bit unsure on how this would work.
Would it be better to check
```php
if ($this instanceof HasBody) {
    return;
}
```

Otherwise we could technically add the trait to the connector, and the interface to the request(s), or vice versa.
I haven't checked if it is, but if it's supported to do that, then I think it should technically work, but might be a bit 'inconsistent', for the lack of a better word.

src/Traits/Body/ChecksForWithBody.php
```diff
@@ -5,7 +5,7 @@
 namespace Saloon\Traits\Body;
 
 use Saloon\Http\PendingRequest;
-use Saloon\Contracts\Body\WithBody;
+use Saloon\Contracts\Body\HasBody;
 use Saloon\Exceptions\BodyException;
 
 trait ChecksForWithBody
```
juse-less: The interface was renamed to HasBody, which it now checks for.

```suggestion
trait ChecksForHasBody
```

src/Traits/Body/ChecksForWithBody.php
```diff
@@ -19,10 +19,10 @@ trait ChecksForWithBody
      */
     public function bootChecksForWithBody(PendingRequest $pendingRequest): void
```
juse-less: The interface was renamed to HasBody, which it now checks for.

```suggestion
    public function bootChecksForHasBody(PendingRequest $pendingRequest): void
```

src/Repositories/Body/ArrayBodyRepository.php
```diff
@@ -57,39 +63,40 @@ public function merge(array ...$arrays): static
...
      *
-     * @param string $key
-     * @return self
+     * @param string|int $key
+     * @param mixed|null $default
+     * @return mixed
      */
-    public function remove(string $key): static
+    public function get(string|int $key, mixed $default = null): mixed
     {
```
juse-less: Since `ArrayBodyRepository` supports adding elements without a key/index, basically acting as a list, would it make sense to also use a nullable key and return the whole body if none was provided?

```suggestion
    public function get(string|int|null $key = null, mixed $default = null): mixed
    {
        if (null === $key) {
            return $this->all();
        }
```

juse-less (commented):
I only had a couple remarks around consistency with the renamed HasBody interface and its corresponding 'checker' trait, as well as 'keying' with the `ArrayBodyRepository`.

Anything else is just suggestions from personal preference.

Something I left out, though, is code like
```php
if (is_array($value)) {
    throw new Exception; // ...
}

$this->value = $value;
```

Which can be a bit simplified using Exceptions as part of expressions (PHP 8.0+)
```php
$this->value = is_array($value) ? $value : throw new Exception; // ...
```

I wasn't entirely sure how widely accepted it is to throw Exceptions as part of expressions.

-----
A bit OffT for the PR, but.. something else I've seen throughout the v2 code base is defining the constructor as part of the interface.
This is more of an implementation detail, rather than a definition, and is generally bad practice, AFAIK.

We only care for the constructor if we're actually instantiating the implementation, and not when we program against the interface (the class is already instantiated in that scenario as well).

In those cases where you need/want to define public properties, as opposed to getters, you can add them to the interface as well, so they are enforced into the implementation.

src/Helpers/Arr.php
Sammyjo20: I think I'll leave this one as this was copied from Laravel's code, and pretty happy with how it works

src/Data/MultipartValue.php
Sammyjo20: In this scenario, I'll keep `!` because it's readable enough

Sammyjo20:
Thank you for the comments @juse-less!

PR #135, opened 2022-12-20
Feature | V2 - Simplified responses and created one single response

Sammyjo20:
This PR simplifies Saloon's responses by removing the requirement for a sender to provide a response. I've also massively simplified the `resolveResponse` logic which before was quite a big method. 

Previously, you could only specify custom responses that extended the base one. You can still do that now, but it now supports your own implementation.

### DIFF
--- a/src/Contracts/Connector.php
+++ b/src/Contracts/Connector.php
@@ -122,9 +122,9 @@ interface Connector
     /**
      * Get the response class
      *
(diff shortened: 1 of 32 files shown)


Sammyjo20 (commented):
Self review

Sammyjo20 (commented):
Self review

PR #136, opened 2022-12-27
Feature | V2 Added headers to body traits

Sammyjo20:
This PR adds default headers to the `FormBodyRepository` and `JsonBodyRepository` classes to depend less on the sender to provide the headers.


Sammyjo20 (commented):
Self review

Sammyjo20 (commented):
Self review

PR #138, opened 2023-01-11
Fix | V2 - Changed the Response class return type to its interface counterpart

juse-less:
I noticed (well.. PhpStorm did..) that the return type was referring to the class which was imported, rather than the interface, which is aliased.

So I changed the return type to use the interface.

### DIFF
--- a/src/Http/PendingRequest.php
+++ b/src/Http/PendingRequest.php
@@ -434,7 +434,7 @@ class PendingRequest implements PendingRequestContract
      *
      * @return \Saloon\Contracts\Response
      */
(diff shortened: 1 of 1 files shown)


Sammyjo20:
Great spot!

juse-less:
Hmm. I'm not sure what the code style complaint is about - some token input it requires.
I'm not sure if it needs a token to pull the image, or some kind of service?

I also saw that the CI is mentioning the usage of Node v12 ->
https://github.com/Sammyjo20/Saloon/actions/runs/3896795880

PR #141, opened 2023-01-21
Feature | V2 Allow other simulated responses to take priority

Sammyjo20:
This PR makes a change to the `DetermineMockResponse` middleware. If another `SimulatedResponsePayload` has already been set on the PendingRequest, the MockClient will be ignored.


Sammyjo20 (commented):
Self review

Sammyjo20:
@juseless Within this PR I have also added the `setAsynchronous` option to the pending request, so the pending request now has a record of if it was sent synchronously or asynchronously! Let me know your thoughts.

Sammyjo20 (commented):
Self review

PR #142, opened 2023-01-22
Feature | V2 Defaults Class

Sammyjo20:
Proof-of-concept idea where we have a defaults class where any application can register default middleware. Could be very useful for Laravel


Sammyjo20 (commented):
Self review

PR #143, opened 2023-01-24
Added and corrected type hints and DocBlocks across the board

juse-less:
I've added and corrected type hints, as well as DocBlocks across the board.

* Some types were missing (either as direct type hint, or in the DocBlock) or was incorrect.
   These should now be correct everywhere.
* Some type hints, as well as DocBlocks, were referencing classes instead of interfaces - changed those to be entirely interfaces.
* I've made all classables in DocBlocks fully qualified, which is partly why so many files has been changed.
* I've also separated some of the more common definitions (like Authenticatable methods) into separate, semantic interfaces, so they can easily be extended, making typing more maintainable.
* I added a base PHPStan config, and updated .editorconfig and .gitignore for that file.
* I also took the liberty to add some more .gitignores while I was adding PHPStan - **let me know if I should revert them**.

All tests still passes, like before.
While I set the PHPStan config to max, it currently only passes on level 5, apart from some of the mocking stuff.
It's on my todo to have a chat to @Sammyjo20 about those, as they were a bit problematic.

A bit later this week, maybe next, I'll sort
* iterable types (PHPStan level 6)
* callable types (PHPStan level 7)
* look into sorting generics for DTOs, among other things

Should I make them as part of this PR, or make new ones for them?
This one is already fairly large.
#120 already tries to adress generics, so maybe do that there?


juse-less:
I think I've solved pretty much everything that was 'wrong' now (missing types, or incorrect DocBlocks).
The files I've excluded from PHPStan (locally) are
* src/Http/Faking/MockClient.php
* src/Traits/Responses/HasSimulationMethods.php

They were tricky to type hint because the requests and/or responses can be many types, and in some cases unsupported types.
So I'll need to speak to @Sammyjo20 about them specifically, to clear things up a bit, so we can correctly add types and make static analysis tools a bit happier. 🙂 
All existing tests (and PHPStan level 5) goes through now.

Sammyjo20 (approved):
This is truly awesome work Juse, I don't have any suggestions just happy to merge! Thank you for your effort!

src/Contracts/Makeable.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Saloon\Contracts;
+
+interface Makeable
+{
+    /**
+     * Instantiate a new class with the arguments.
+     *
+     * @param mixed ...$arguments
+     * @return static
```
sneycampos, line 13: @juse-less why are you returning static here but in ArrayStore you uses $this?

juse-less: Depending on what you're referring to -

The `make()` method(s) returns a new instance, whereas calls to objects returns _the same_ instance.
`static` is the same class, whereas `$this` is the very same instance you call the method on.

The native type hint is still `static`, however, because PHP doesn't support type hinting `$this`.

```php
class MyClass
{
    public static function make(): static
    {
        return new static;
    }

    public function setValue(mixed $value): static
    {
        // ... set value, etc ...

        return $this;
    }
}
```

sneycampos: Sorry @juse-less, i mean in phpdoc block only

juse-less: I'm guessing you're referring to this method (as an example), and the others like it:
https://github.com/Sammyjo20/Saloon/blob/7cb97128732c6e1f7cf9c75f20e173922e3005d7/src/Contracts/ArrayStore.php#L29-L31

The reason is that the method returns the same instance, and not a new instance.
If, for example, the class would be immutable, we'd return `new static` instead of `$this`, and therefore also type hint with `@return static`.

sneycampos: Got it! Thank your for your explanation and your contribution!

PR #144, opened 2023-01-24
Feature | V2 Better Merging Pipeline

Sammyjo20:
This PR makes some more improvements to the middleware pipelines, specifically switching to a `Pipe` class for each pipe which now contains the name. This means I no longer need a separate array for the pipe names and when merging pipelines, all the names can be brought across. 

The default fixture directory will now be created even when it doesn't exist. @simple-hacker @Gummibeer 

Todo: Throw an exception if merging results in two of the same middleware?

### DIFF
--- a/src/Helpers/Pipeline.php
+++ b/src/Helpers/Pipeline.php
@@ -4,9 +4,9 @@ declare(strict_types=1);
 
 namespace Saloon\Helpers;
 
-use Saloon\Contracts\Pipeline as PipelineContract;
 use Saloon\Data\Pipe;
 use Saloon\Exceptions\DuplicatePipeNameException;
+use Saloon\Contracts\Pipeline as PipelineContract;
 
 class Pipeline implements PipelineContract
 {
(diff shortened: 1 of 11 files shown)


Sammyjo20 (commented):
Self review

src/Helpers/Pipeline.php
```diff
@@ -77,10 +72,21 @@ public function setPipes(array $pipes): static
...
+
+    /**
+     * Check if a given pipe exists for a name
+     *
+     * @param string $name
+     * @return bool
+     */
+    protected function pipeExists(string $name): bool
+    {
+        return ! empty(array_filter($this->pipes, static fn (Pipe $pipe) => $pipe->name === $name));
```
juse-less: I'd change to a foreach here, so you can bail immediately.
If someone has many pipes in the pipeline, it can drastically improve performance.
We really only care if it has one, so there's no need to continue iterating.

```suggestion
        foreach ($this->pipes as $pipe) {
            if ($pipe->name === $name) {
                return true;
            }
        }
        
        return false;
```

Sammyjo20: Awesome, thanks @juse-less !

Sammyjo20 (commented):
Self review

PR #145, opened 2023-01-25
Feature | V2 Iterable type hints

juse-less:
I've added type hints to all iterables, except for
* src/Contracts/MockClient.php
* src/Http/Faking/MockClient.php
* src/Traits/Macroable.php

All three has tricky types, so doing them last when everything else is completed.

I also changed the constructor on ArrayBodyRepository to use an array directly instead of mixed, as it's not bound to mixed.
As part of that I also adjusted its test that tests if it correctly throws an Exception if the input is not an array.
I noticed that two of the tests had their labels mixed up (but the actual tests were correct).
So I also switched the tests content, to match their test label.

I also widened the `Response::json()` method to accept int, and not only string|null.
This is the same type that `Response::collect()` has, as well.
I also know that some APIs will return a collection of items, without being wrapped.
That way you can grab an item at a specific index directly, if you need to.

I've also widened the Generator type to iterable in the Pool.
We weren't using the benefit of a Generator, and were entirely using the iterable feature of it.
Do note that `Pool::getRequests()` return type was also changed, as part of that.
Should a Generator still be needed, it's easily sortable by manually yielding the iterable.

PHPStan now passes on level 6, except from my currently ignored local files (same as the initial list above).
All tests still also pass.

### DIFF
--- a/src/Contracts/ArrayStore.php
+++ b/src/Contracts/ArrayStore.php
@@ -9,7 +9,7 @@ interface ArrayStore
     /**
      * Retrieve all the items.
      *
-     * @return array
+     * @return array<string, mixed>
      */
     public function all(): array;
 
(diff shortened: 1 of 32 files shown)


Sammyjo20 (approved):
Another great PR, thank you Juse.

PR #146, opened 2023-02-01
Feature | V2 Retry Functionality

Sammyjo20:
This PR introduces a new exciting method into Saloon... `sendAndRetry`! This method will automatically send a request again if it fails. I've taken a lot of inspiration from Laravel's HTTP client features so the features are almost identical.

You can:
- Specify maximum attempts
- Specify an interval in milliseconds 
- Specify a callable to handle if we should retry or not
- Specify if we should throw on the final error or return a response

This PR also adds a new `hasRequestFailed()` method on the request and connector which allows you to further control what a `failed` response is.

This PR also adds a new method `dtoOrFail` and changes the existing `dto` method to work even if the response has failed.


Sammyjo20 (commented):
Self review

PR #149, opened 2023-02-02
WIP: Feature | V2 Callable type hints

juse-less:
I've added type hints to all callable types, except for test-related classes, interfaces, etc.

I'm not yet quite sure how to solve a couple of issues I'm facing, hence marked as draft/WIP.

<details>
<summary>Conditionable trait</summary>

I believe the issue is that we're checking if the value is truey, respectively falsey.
Because only the given values (array|float|int|string|bool|null) can be either truey (non-empty array, 0.0, 0, non-empty string, or true), or falsey (empty array, != 0.0, != 0, empty string, or false), I believe that's why it's complaining.
As in - we've checked either falsey or truey, so that's the implied value now (not object, etc).
```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   Traits/Conditionable.php (in context of class <any classable using the trait>)
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------
  32     Parameter #2 $ of callable callable($this(<any classable using the trait>), TValue): void expects TValue, array|float|int|string|false|null given.
         ✏️  Traits/Conditionable.php
  53     Parameter #2 $ of callable callable($this(<any classable using the trait>), TValue): void expects TValue, array|float|int|string|false|null given.
         ✏️  Traits/Conditionable.php
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
</details>

<details>
<summary>PendingRequest</summary>

The issue here is a bit the same as to why I've been avoiding to fix test-related things so far.
The DetermineMockResponse will take a PendingRequest, but alter it to be MockResponse (if it's mocked).
The MockResponse is a SimulatedResponsePayload, but that isn't the same as a PendingResponse.
So the Response type is actually modified.
This makes it a bit problematic to be _type safe_, even if we only ever type with MockResponse in tests.
Nonetheless, in the case of DetermineMockResponse, it returns either PendingRequest, or, if it's mocked, MockResponse.
And thereby should also be able to retrieve a MockResponse.

Anything in the request pipeline should be able to accept a MockResponse and not only a PendingRequest, or they will fail when mocking.
```
------ -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   Http/PendingRequest.php
 ------ -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  272    Parameter #1 $callable of method Saloon\Contracts\MiddlewarePipeline::onRequest() expects callable(Saloon\Contracts\PendingRequest|Saloon\Http\Faking\MockResponse):
         Saloon\Contracts\PendingRequest|Saloon\Http\Faking\MockResponse|void, Saloon\Http\Middleware\DetermineMockResponse given.
         ✏️  Http/PendingRequest.php
 ------ -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
</details>


juse-less:
Forgot to select the v2 branch. 🤦‍♀️

PR #150, opened 2023-02-02
Feature | V2 Callable type hints

juse-less:
I've added type hints to all callable types, except for test-related classes, interfaces, etc.

I'm not yet quite sure how to solve a couple of issues I'm facing, hence marked as draft/WIP.

<details>
<summary>Conditionable trait</summary>

I believe the issue is that we're checking if the value is truey, respectively falsey.
Because only the given values (array|float|int|string|bool|null) can be either truey (non-empty array, 0.0, 0, non-empty string, or true), or falsey (empty array, != 0.0, != 0, empty string, or false), I believe that's why it's complaining.
As in - we've checked either falsey or truey, so that's the implied value now (not object, etc).
```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   Traits/Conditionable.php (in context of class <any classable using the trait>)
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------
  32     Parameter #2 $ of callable callable($this(<any classable using the trait>), TValue): void expects TValue, array|float|int|string|false|null given.
         ✏️  Traits/Conditionable.php
  53     Parameter #2 $ of callable callable($this(<any classable using the trait>), TValue): void expects TValue, array|float|int|string|false|null given.
         ✏️  Traits/Conditionable.php
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
</details>

<details>
<summary>PendingRequest</summary>

The issue here is a bit the same as to why I've been avoiding to fix test-related things so far.
The DetermineMockResponse will take a PendingRequest, but alter it to be MockResponse (if it's mocked).
The MockResponse is a SimulatedResponsePayload, but that isn't the same as a PendingResponse.
So the Response type is actually modified.
This makes it a bit problematic to be _type safe_, even if we only ever type with MockResponse in tests.
Nonetheless, in the case of DetermineMockResponse, it returns either PendingRequest, or, if it's mocked, MockResponse.
And thereby should also be able to retrieve a MockResponse.

Anything in the request pipeline should be able to accept a MockResponse and not only a PendingRequest, or they will fail when mocking.
```
------ -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   Http/PendingRequest.php
 ------ -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  272    Parameter #1 $callable of method Saloon\Contracts\MiddlewarePipeline::onRequest() expects callable(Saloon\Contracts\PendingRequest|Saloon\Http\Faking\MockResponse):
         Saloon\Contracts\PendingRequest|Saloon\Http\Faking\MockResponse|void, Saloon\Http\Middleware\DetermineMockResponse given.
         ✏️  Http/PendingRequest.php
 ------ -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
</details>

### DIFF
--- a/src/Http/Pool.php
+++ b/src/Http/Pool.php
@@ -12,9 +12,6 @@ use GuzzleHttp\Promise\PromiseInterface;
 use Saloon\Contracts\Pool as PoolContract;
 use Saloon\Exceptions\InvalidPoolItemException;
 
-use function is_callable;
-use function is_null;
-
 class Pool implements PoolContract
 {
     /**
--- a/src/Http/Pool.php
+++ b/src/Http/Pool.php
@@ -24,14 +24,14 @@ class Pool implements PoolContract
     /**
      * Handle Response Callback
      *
-     * @var (\Closure(\Saloon\Contracts\Response, array-key $key, \GuzzleHttp\Promise\PromiseInterface $poolAggregate): void)|null
+     * @var \Closure(\Saloon\Contracts\Response, array-key $key, \GuzzleHttp\Promise\PromiseInterface $poolAggregate): (void)|null
      */
     protected ?Closure $responseHandler = null;
 
     /**
      * Handle Exception Callback
      *
-     * @var (\Closure(mixed $reason, array-key $key, \GuzzleHttp\Promise\PromiseInterface $poolAggregate): void)|null
+     * @var \Closure(mixed $reason, array-key $key, \GuzzleHttp\Promise\PromiseInterface $poolAggregate): (void)|null
      */
     protected ?Closure $exceptionHandler = null;
 
--- a/src/Http/Pool.php
+++ b/src/Http/Pool.php
@@ -12,6 +12,9 @@ use GuzzleHttp\Promise\PromiseInterface;
 use Saloon\Contracts\Pool as PoolContract;
 use Saloon\Exceptions\InvalidPoolItemException;
 
+use function is_callable;
+use function is_null;
+
 class Pool implements PoolContract
 {
     /**
@@ -24,14 +27,14 @@ class Pool implements PoolContract
     /**
      * Handle Response Callback
      *
-     * @var \Closure|null
+     * @var (\Closure(\Saloon\Contracts\Response, array-key $key, \GuzzleHttp\Promise\PromiseInterface $poolAggregate): void)|null
      */
     protected ?Closure $responseHandler = null;
 
     /**
      * Handle Exception Callback
      *
-     * @var \Closure|null
+     * @var (\Closure(mixed $reason, array-key $key, \GuzzleHttp\Promise\PromiseInterface $poolAggregate): void)|null
      */
     protected ?Closure $exceptionHandler = null;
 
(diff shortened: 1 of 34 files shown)


juse-less:
@Sammyjo20 I've been waiting with bringing the test-related problem(s) up, so it could be resolved at once.
I think it'd be good to address the problem of type safety with mocks.
Have a look at my comments for the PendingRequest part in this PR, for more info.

src/Http/Pool.php
```diff
@@ -12,6 +12,9 @@
 use Saloon\Contracts\Pool as PoolContract;
 use Saloon\Exceptions\InvalidPoolItemException;
 
+use function is_callable;
+use function is_null;
```
Sammyjo20: Is there a benefit for explicitly defining these?

src/Http/Pool.php
```diff
@@ -47,71 +50,77 @@ class Pool implements PoolContract
...
      * Specify a callback to happen for each successful request
      *
-     * @param callable $callable
+     * @param callable(\Saloon\Contracts\Response, array-key $key, \GuzzleHttp\Promise\PromiseInterface $poolAggregate): (void) $callable
      * @return $this
      */
     public function withResponseHandler(callable $callable): static
     {
-        $this->responseHandler = $callable;
+        $this->responseHandler = $callable(...);
```
Sammyjo20, line 86: What does the ellipses do here?

Sammyjo20 (changes requested):

src/Http/Pool.php
juse-less: It's the first-class callable syntax.
They turn any callable into a Closure, in the same fashion as Closure::fromCallable().
It actually worked without, but is technically wrong as a Closure is an invokable class (callable), but a callable isn't necessarily a Closure.
So changed it to explicitly turn them into Closures, to fit the type hint of the property.

The syntax was added in PHP 8.1, which is the lowest version in Saloon v2, per composer.json.

https://www.php.net/manual/en/functions.first_class_callable_syntax.php

src/Http/Pool.php
juse-less: I'm not entirely sure. Possibly negligible performance benefits, because of how PHP searches for namespaces when it tries to resolve them.

But in this case, it was my PhpStorm automagically adding them. I didn't realise it.
I'll search for any function import and remove them. 👍

Sammyjo20:
I'm really not sure what to do with the conditionable trait, I copied it from Laravel's source code and I'm not entirely convinced if we need it on things, but I added it anyway for convenience. 

With the PendingRequest, do you think we should allow a middleware to also accept a SimulatedResponsePayload? It was more for convenience that I made it so you could return either nothing, and it'll carry on using the pending request, the actual pending request instance or an instance of the simulated response payload, but the pipeline itself will always have a pending request being passed down. Perhaps one to ignore? Not sure what to do. Let me know if you find a better name for `SimulatedResponsePayload` because I'm not a fan 😆

src/Http/Pool.php
Sammyjo20: You learn things every day, thanks Juse!

juse-less:
Regarding the `Conditionable` trait, it's at least not used internally.
I don't personally use it, which I otherwise usually do frequently.
I'm not sure about others, though.

Technically speaking, we could add the specific error as an ignore in PHPStan.

Regarding the PendingRequest and mocking, I'm honestly not sure how mocks are usually made.

Spontaneously, I feel like a response should be a response, regardless if it's mock data or not.
So I'd probably create a MockResponse, based on a real Response, but with mock-specific stuff added.
Then fill in the needed mock-data on the MockResponse.

I'm not sure if that's correct or not, though. Nor if it's easily possible, as I haven't gone deep into Saloons testing stuff yet.

If we'd go that route, however, it'd solve the `SimulatedResponsePayload` name by simply remove the class \*hint\* \*hint\* 😉

Sammyjo20:
Let me have a think about that one, the reason why I went with SimulatedResponsePayload is that it had to be used for the caching plugin too. Perhaps something like `FakeResponseData`?

Or if we did go down the route `FakeResponse` could have a method like `toResponse` like it already has. I'd rather it not extend `Response` because then you would lose `MockResponse::make('data', 'status', 'headers')

FakeResponse would be a simple rename from SimulatedResponsePayload.

Sammyjo20:
Okay, also what if it's just fine as it is for V2, Maybe enhancing all that could be for v3? Just thinking if a lot of this is just starting to distract me from releasing v2.

juse-less:
> Okay, also what if it's just fine as it is for V2, Maybe enhancing all that could be for v3? Just thinking if a lot of this is just starting to distract me from releasing v2.

Yeah. I agree.
I'll just silence the error in PHPStan for now?

Basically - we know we'll receive a MockResponse, those times it'll be returned (testing).
Just document it, as others middlewares could have issues if they only accept PendingRequest, and then receive the MockResponse.

It's a tricky one. I did take a quick stab at making a FakeResponse/MockResponse that extends Response, which tried to resolve
> I'd rather it not extend `Response` because then you would lose `MockResponse::make('data', 'status', 'headers')`
But it was tricky forming all the needed parts (like Connector), based on the `MockResponse::make()` arguments.

So.. for this PR then:
1. Silence the Conditionable trait errors for now, as they are not technically errors in this case (imo).
1. Silence the `PendingRequest|MockResponse` error.

Sammyjo20:
Maybe I should still rename it to something like `FakeResponseData` to be a bit snappier?

Also I don't think that's quite right, this is the code in MiddlewarePipeline.php, as you can see I detect what has been returned back:

```php
public function onRequest(callable $callable, bool $prepend = false, ?string $name = null): static
    {
        $this->requestPipeline = $this->requestPipeline->pipe(function (PendingRequest $pendingRequest) use ($callable) {
            $result = $callable($pendingRequest);

            if ($result instanceof PendingRequest) {
                return $result;
            }

            if ($result instanceof SimulatedResponsePayload) {
                $pendingRequest->setSimulatedResponsePayload($result);
            }

            return $pendingRequest;
        }, $prepend, $name);

        return $this;
    }
```

So I'd just need to document that even if you return a fake response it will not give you a fake response in the next request middleware. Very small thing but useful to know anyway.

Sammyjo20:
> I'll just silence the error in PHPStan for now? 

Yes please

juse-less:
> > I'll just silence the error in PHPStan for now?
> 
> Yes please

Ahh! I didn't reflect on that. That eases things. 🙂
Then it's slightly less dramatic than my initial interpretation. 👍

juse-less:
@Sammyjo20 With your pointer on `PendingRequest` always being returned, regardless of `$result` being a `SimulatedResponsePayload` or not, I was able to sort it without silencing the error.

I generated a baseline, so future CI can run without not-yet-fixed errors.
Note, however, that one of the errors can't be ignored, but should be fixed by your work on #151.

I also realised that the Pipe class accepted `mixed` for the `$callable` parameter, which I changed to `callable`.
https://github.com/Sammyjo20/Saloon/pull/150/commits/7327c1139f20a704f38fc5b793fc7840ef2f48de

And I also sorted some non-callable stray-errors.

The PR should now be ready for review. 🙂

juse-less:
I just realised GitHub says the PR is blocked from merging until your (@Sammyjo20) change requests are resolved.
But.. the discussions (where the change requests were made) above are all resolved, so I'm not sure what it's talking about..

How do I make the PR acknowledge all discussions are resolved? 🤔

Sammyjo20 (approved):
Great work thank you!

Sammyjo20:
@juse-less Would you be able to catch this branch up to v2 please, I fixed an issue with tests

juse-less:
@Sammyjo20 I think it should be up to date now.

Can you just have a second look, so that everything looks ok?
Looks like I accidentally made a merge with upstream, instead of a rebase.
As part of that process, some things weren't correctly re-applied.
So I applied them again afterwards.

Let me know if I should just re-do the whole thing.
At least PHPStan goes through.
All tests goes through, except for one that's being skipped
- Tests\Feature\RetryRequestTest
   - a failed request can have an interval between each attempt → Occasionally fails

Sammyjo20 (approved):

PR #151, opened 2023-02-02
Fix | Improve sendAndRetry

Sammyjo20:
I've also added a PHPStan Github action 😁

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -1,0 +1,26 @@
+name: PHPStan
+
+on:
+  push:
+    paths:
+      - '**.php'
+      - 'phpstan.neon.dist'
+
+jobs:
+  phpstan:
+    name: phpstan
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.1'
+          coverage: none
+
+      - name: Install composer dependencies
+        uses: ramsey/composer-install@v2
+
+      - name: Run PHPStan
+        run: ./vendor/bin/phpstan --error-format=github
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -23,4 +23,4 @@ jobs:
         uses: ramsey/composer-install@v2
 
       - name: Run PHPStan
-        run: ./vendor/bin/phpstan --error-format=github
+        run: ./vendor/bin/phpstan analyse --error-format=github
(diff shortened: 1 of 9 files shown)


Sammyjo20 (commented):
Self review

juse-less:
Regarding the PHPStan reports, it's because I've set the level to max in the default PHPStan config (phpstan.dist.neon).

I intended on fixing everything before release (and the GitHub action).
In the meantime I guess we can set the level to 6, and I can increase it in each PR I add for type hints.
This also makes a lot of sense.
If you set it to 6, I'll bump it to 7 in my callable typehint PR. 🙂

Sammyjo20 (commented):
Self review

src/Contracts/MockClient.php
```diff
@@ -104,7 +104,7 @@ public function assertNotSent(string|callable $request): void;
      * Assert JSON data was sent
      *
      * @param string $request
-     * @param array $data
+     * @param array<mixed, mixed> $data
```
juse-less, line 107: Keys can only really be `int|string` *1. To type hint array keys, you can use `array-key` for those.

1. unless you use WeakMaps, at which point objects are turned into internal hashes.

juse-less: Keys can only really be `int|string` *1. To type hint array keys, you can use `array-key` for those.

1. unless you use `\WeakMaps`, at which point objects are turned into internal hashes.

```suggestion
     * @param array<array-key, mixed> $data
```

composer.json
```diff
@@ -64,6 +64,9 @@
         ],
         "test": [
             "./vendor/bin/pest"
+        ],
+        "pstan": [
+            "./vendor/bin/phpstan"
```
juse-less: I think it's better to be explicit in CI:s, so maybe add `analyse` here.
Replace with `analyze` depending on English/American (PHPStan supports both). 😅 

```suggestion
            "./vendor/bin/phpstan analyse"
```

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,26 @@
...
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.1'
+          coverage: none
+
+      - name: Install composer dependencies
+        uses: ramsey/composer-install@v2
+
+      - name: Run PHPStan
+        run: ./vendor/bin/phpstan --error-format=github
```
juse-less: I think it's better to be explicit in CI:s, so maybe add `analyse` here.
Replace with `analyze` depending on English/American (PHPStan supports both). 😅 

```suggestion
        run: ./vendor/bin/phpstan analyse --error-format=github
```

Sammyjo20 (commented):
Self review

PR #152, opened 2023-02-04
Feature | V2 Paginator

Sammyjo20:


src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,184 @@
...
+namespace Saloon\Http\Paginators;
+
+use Closure;
+use GuzzleHttp\Promise\PromiseInterface;
+use Saloon\Contracts\Connector;
+use Saloon\Contracts\Request;
+use Saloon\Contracts\Response;
+
+// TODO 1: Look into serialising the Connector and original Request,
+//           to ensure that we can rebuild the paginator state without storing the entire multiverse.
```
Sammyjo20: 🤣

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,184 @@
...
+// TODO 2: Make it easier to extend the RequestPaginator. Preferably via callbacks, so we ideally don't even need separate classes.
```
Sammyjo20: I think it'll be cool and very Saloon-like if we allow people to create their own "Paginator" classes that extend our base paginator. When they extend the base abstract paginator, they will be required to define the missing methods like:

- isOnLastPage 
- getResults
- getTotal

That way people could have a PokeapiPaginator, ForgePaginator, then inside of our request `paginate` method they would just return `new PokeapiPaginator`?

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,184 @@
...
+// TODO 3: Because both page-based and offset-based pagination just bumps the 'paging' number,
+//           would it make sense to do all pagination in _the_ RequestPaginator, but allow to specify the counter somehow?
+//         Maybe a callback that receives a copy of the original Request, as well as the latest Response (which has the corresponding latest Request and PendingRequest),
+//           and have that callback set the next 'page' on the new Request?
```
Sammyjo20: You can get the original request from the pending request, but I imagine the pending request _should_ have all the correct request properties in like body, query parameters etc.

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,184 @@
...
+    /**
+     * @return ($this->async is true ? \GuzzleHttp\Promise\PromiseInterface : TResponse)
+     */
+    public function current(): Response|PromiseInterface
+    {
+        $request = clone $this->originalRequest;
+        $request->query()->merge([
+            'limit', $this->limit,
+            'offset', $this->offset,
+        ]);
```
Sammyjo20: I wonder if it would be better to create a pending request and modify that instead, so the original request is left untouched:

```suggestion
        $pendingRequest = $this->connector->createPendingRequest(clone $this->originalRequest);
        $pendingRequest->query()->merge([
            'limit', $this->limit,
            'offset', $this->offset,
        ]);
```

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,184 @@
...
+
+        // TODO: async
+
+        return $this->lastResponse = $this->connector->send($request);
```
Sammyjo20:
```suggestion
        return $this->lastResponse = $pendingRequest->send();
```

src/Http/Paginators/RequestPaginator.php
```diff
@@ -0,0 +1,152 @@
...
+     * @return array{
+     *     connector: \Saloon\Contracts\Connector,
+     *     original_request: \Saloon\Contracts\Request,
+     *     limit: int|null,
+     * }
+     */
+    public function __serialize(): array
+    {
+        // TODO: figure out how to serialise the 'has next page' resolver/callback.
+
```
Sammyjo20: If we switched to making the user extend their own class, we wouldn't have to worry about serializing closures

src/Http/Paginators/ResponsePaginator.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * @return void
+     */
+    public function previous(): void
+    {
+        $this->index--;
+    }
+}
```
Sammyjo20: This looks interesting, although I can't see where it's being used

src/Contracts/Connector.php
```diff
@@ -173,6 +173,14 @@ public function hasMockClient(): bool;
      */
     public function pool(iterable|callable $requests = [], int|callable $concurrency = 5, callable|null $responseHandler = null, callable|null $exceptionHandler = null): Pool;
 
+    /**
+     * @param \Saloon\Contracts\Request $request
+     * @param mixed ...$additionalArguments
+     *
+     * @return \Saloon\Contracts\RequestPaginator
+     */
+    public function paginate(Request $request, mixed ...$additionalArguments): RequestPaginator;
```
Sammyjo20: What do you imagine we could use the $additionalArguments for?

Sammyjo20 (commented):
Looking really good so far

Sammyjo20:
Also if you want to try it against Saloon's test API here are two endpoints:

per-page/page pagination: https://tests.saloon.dev/api/superheroes/per-page
offset-limit pagination: https://tests.saloon.dev/api/superheroes/limit-offset

src/Contracts/Connector.php
juse-less: In case the constructor on a paginator wants more arguments.
Like.. if you want to set the limit.
However, I'm most likely removing this in favour of your other idea on extending paginators in Saloon fashion, which would defeat this purpose.
At least in general.

src/Http/Paginators/ResponsePaginator.php
juse-less: https://github.com/juse-less/Saloon/commit/0aae49bacc4e5f2a4bdc9925bb40b74f00631033

There was already some info on it, but added DocBlock examples.
For some reason PhpStorm won't show previews for me, so if someone is able to check out my branch and see so that the example renders correctly in DocBlocks, that'd be very helpful.

src/Http/Paginators/OffsetRequestPaginator.php
juse-less: We never alter the original request - we clone it all the time.
This is to keep tabs on the literal initial request, which is the same regardless of amount of requests made.
Then we can follow the regular 'send' flow, by simply bumping pages.

I guess, technically speaking, the current and suggested ways are semantically equivalent.

src/Http/Paginators/OffsetRequestPaginator.php
juse-less: That is very true. And, in theory, the PendingRequest should already have had middlewares applied, which would ideally help with the serialisation situation.

Although.. it'd probably be better to always run the middlewares, in case something is modified depending on page number, or something.

src/Http/Paginators/OffsetRequestPaginator.php
juse-less: That makes perfect sense. I like the idea and significantly reduces the complexity and issues I had trying to figure out a decent way to handle it, without causing poor DX. 🙂

src/Http/Paginators/ResponsePaginator.php
juse-less: Now that I think of it..
The `RequestPaginator::valid()` method always checks if the _next_ page is valid, not previous.
So this would cause issues.

I'd say it's highly unlikely anyone would use it anyway, so I'll just remove it altogether.
If a reverse iterator is needed, it can easily be created separately.

src/Http/Paginators/OffsetRequestPaginator.php
Sammyjo20: Everything is run on the PendingRequest’s constructor, so once you have the instance, it has already been through all plugins and middleware and executed them, the pending request in this instance is completely ready to be sent and all the properties should be merged into it.

juse-less:
I'll fix the failing tests when I move 'resolving' parts out of `PageRequestPaginator` and `OffsetRequestPaginator` and into the test implementations.

I still have a lot of stuff to figure out. I think each of them should have an 'attached' `TODO` next to them.

src/Http/Paginators/ResponsePaginator.php
juse-less: I went through with it, and removed the `previous()` methods.

src/Http/Paginators/RequestPaginator.php
juse-less: I have added the `hasNextPage()` method, which needs to be implemented.
There's a TODO on it, to figure out a better name, though.
`isFinished()`? Or something like it, to better fit offset-based paging.

src/Http/Paginators/OffsetRequestPaginator.php
juse-less: I have implemented it this way, with a base `RequestPaginator`.

I also took the liberty to implement a base `PageRequestPaginator` and `OffsetRequestPaginator`.
Each of which extends the `RequestPaginator`, but implements some of the resolving methods, based on common names (offset, limit, page, etc).

Still WIP, with plenty of TODO to figure out, though.

juse-less:
> Also if you want to try it against Saloon's test API here are two endpoints:
> 
> per-page/page pagination: https://tests.saloon.dev/api/superheroes/per-page offset-limit pagination: https://tests.saloon.dev/api/superheroes/limit-offset

@Sammyjo20 Technically speaking, 'paged' APIs should really always have 1 as the first page.
But noticed that the per-page endpoint doesn't return any first page - only last.
I'm not sure if it makes sense to add it for consistency, even though it'll always be 1?

juse-less:
Regarding serialising the `RequestPaginator`s, I think it's better to add that as a feature request after v2 has landed.
That one could take quite a bit of time to figure out.

I believe the only blocker, technically speaking, are middlewares as they can be Closures, that aren't really serialisable, AFAIK.
There are packages to make it easier, but fairly certain it doesn't always work.

`Request` headers, query parameters, body payload, and configuration should all be easily serialisable.

Same thing with the `Connector`. The only caveat is if people add things like credentials as arguments.
But considering both the `Request` and `Connector` should serialise themselves, it's more of a documentation thing for implementors to add those properties to the `__serialisation()` and `__unserialisation()` methods.
So not really a problem.

-----
Maybe we could ask how much it'd be used, as well.
If it wouldn't really be used, then we can drop the idea so we don't need to maintain something no one would use, as it feels kinda complex to implement.

Context:
We've been talking about making `RequestPaginator`s serialisable so that they can be sent into things like Laravels background Jobs.
This makes it possible to iterate, say, 25 requests, and pass the iterators to another Job and continue there, automagically.
In order to retain the same state, we need to rebuild the same Connector and Request, so that they send the very same data, apart from whatever is changed per request.
So it's not as easy as passing on current page, or so, as Request or Connector could change between Jobs.

Sammyjo20:
> I think it's better to add that as a feature request after v2 has landed

Yeah I actually agree on this one. We firstly have the issue as you mentioned with middleware and while I don't mind telling people "if you want to serialize the paginator, you have to switch to invokable class middleware" I don't actually know how many people would really use it. The only time I'd use it, would be if I was rate limited by the API - I would want to queue a job with the backoff time and then run the process again. 

That being said, this would actually solve a really annoying issue we have at work where an API only gets so far and then stops at exactly the same point because it gets rate limited. Did you figure out a way to "resume" the iteration process from where it was left off? I wonder if we set the page in the "rewind" method to be the unserialized last page, so it starts at page 10 for example. What we also could do is when we know we are on the last page, we reset the "starting point" back to zero so if someone starts iterating over it again, the rewind doesn't start at page 10 again. 💡

I kind of think it'll be worth it, especially when all they have to do is migrate middleware to be invokable classes which are cleaner typically. If we can implement it later on in v2  hopefully without any breaking changes then happy days, but I think what you have come up with will be great for now.

juse-less:
> Did you figure out a way to "resume" the iteration process from where it was left off?

Yeah - I've solved the iteration/paging continuation.
Have a look at the the [`rewind()`](https://github.com/juse-less/Saloon/blob/d257cdf35d4a3334f6dbb0ee98400217a0b679ed/src/Http/Paginators/RequestPaginator.php#L192-L194) method.
See this [DocBlock](https://github.com/juse-less/Saloon/blob/d257cdf35d4a3334f6dbb0ee98400217a0b679ed/src/Contracts/RequestPaginator.php#L41-L54) in the contract for more info.
By default, we'll continue where it left off. But can be disabled.

If we want to continue in a separate process, we'd need to be able to serialise- and unserialise it to retain the same state.

> If we can implement it later on in v2 hopefully without any breaking changes then happy days, but I think what you have come up with will be great for now.

Currently, the serialisation bits are actually a separate interface.
So what we'd do is remove it from this feature branch, and add it later, with a trait, or so, that can be added to custom Paginators for easy serialisation.
So I'm pretty sure that we'd be able to add the serialisation in a separate feature PR, without breaking changes.

But let me have a think next week, and maybe I can solve it with non-Closure middlewares.

Sammyjo20:
I have used the Opis/Closure package before which allows you to serialize closures. If that's the only thing holding it back I would actually want the ability to serialize it in v2, if you agree? Otherwise we could suggest people install the serializable closures package to be able to use the feature, much like the `collect()` method on the response requiring the Illuminate/Support package?

https://github.com/opis/closure

Sammyjo20:
Actually we would use `laravel/serializable-closure` as it is a fork that works with Laravel. I believe Taylor wrote why Laravel is using that one

src/Contracts/Connector.php
Sammyjo20: Sounds good!

src/Contracts/RequestPaginator.php
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * @param callable(int $pendingRequests): (int)|int $concurrency
+     * @param callable(TResponse $response, array-key $key, \GuzzleHttp\Promise\PromiseInterface $poolAggregate): (void)|null $responseHandler
+     * @param callable(mixed $reason, array-key $key, \GuzzleHttp\Promise\PromiseInterface $poolAggregate): (void)|null $exceptionHandler
+     */
+    public function pool(
+        callable|int $concurrency = 5,
+        ?callable $responseHandler = null,
+        ?callable $exceptionHandler = null,
+    ): Pool;
```
Sammyjo20: I still think this is genius, such a cool way to implement pools + pagination

src/Contracts/RequestPaginator.php
```diff
@@ -0,0 +1,191 @@
...
+     * I.e., if you go through 2 iterations in a loop, then break out of it, and iterate again in a new loop, it'll be rewound to the first element.
+     * This makes sure that the rewinding step is skipped.
+     *
+     * @param bool $ignoreRewinding
+     *
+     * @return $this
+     *
+     * @TODO Come up with a better name for this method.
+     */
+    public function ignoreRewinding(bool $ignoreRewinding = true): static;
```
Sammyjo20: I like `ignoreRewinding`, you could call it `preventRewinding`?

Sammyjo20: Very clever way of doing it though!

src/Contracts/RequestPaginator.php
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * Returns whether or not the Paginator will continue where it left off in a previous loop, or not.
+     *
+     * @return bool
+     *
+     * @see \Saloon\Contracts\RequestPaginator::ignoreRewinding()
+     *
+     * @TODO Come up with a better name for this method.
+     */
+    public function shouldRewind(): bool;
```
Sammyjo20: I think `shouldRewind` is fine but you could call it `isRewindingEnabled`

src/Contracts/RequestPaginator.php
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * @return int|null
+     *
+     * @TODO: Does all type of paging actually have limit, or should this be extracted somewhere else?
+     */
+    public function limit(): ?int;
```
Sammyjo20: I imagine it would, but most APIs have defaults

src/Http/Paginators/OffsetRequestPaginator.php
Sammyjo20: I also think we should make sure we can let people customise the query parameters that are set, nothing fancy maybe properties or a method they extend, but I once dealt with an API that used the terminology "top" and "skip" instead of "limit" and "offset" 🤮

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,294 @@
...
+        return $this->lastPage();
+    }
+
+    /**
+     * @return int
+     */
+    public function firstPage(): int
+    {
+        // TODO: Or should we use something from the response?
+        return 1;
```
Sammyjo20: I think this is fine, but allow people to extend it, which looks like people can

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    /**
+     * @return int
+     */
+    public function firstOffset(): int
+    {
+        // TODO: Or should we use something from the response?
+        return 0;
```
Sammyjo20: Again, I think this is fine to get people to define

Sammyjo20: Additional logic would be required to "guess" it so I think keep it simple

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * @return int|null
+     */
+    public function previousPage(): ?int
+    {
+        $page = $this->currentPage();
+
+        return $page > $this->firstPage()
+            ? $page - 1
+            : null;
```
Sammyjo20:
```suggestion
        return $page > $this->firstPage() ? $page - 1  : null;
```

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * @return int|null
+     */
+    public function previousOffset(): ?int
+    {
+        $previousOffset = $this->currentOffset() - $this->limit();
+
+        return $previousOffset > $this->firstOffset()
+            ? $previousOffset
+            : null;
```
Sammyjo20:
```suggestion
        return $previousOffset > $this->firstOffset() ? $previousOffset : null;
```

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * @return int|null
+     */
+    public function nextPage(): ?int
+    {
+        $page = $this->currentPage();
+
+        return $page < $this->lastPage()
+            ? $page + 1
+            : null;
```
Sammyjo20:
```suggestion
        return $page < $this->lastPage() ? $page + 1 : null;
```

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * @return int|null
+     */
+    public function nextOffset(): ?int
+    {
+        $nextOffset = $this->currentOffset() + $this->limit();
+
+        return $nextOffset < $this->lastOffset()
+            ? $nextOffset
+            : null;
```
Sammyjo20:
```suggestion
        return $nextOffset < $this->lastOffset() ? $nextOffset : null;
```

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,294 @@
...
+    {
+        return $this->totalEntries() - $this->limit();
+    }
+
+    protected function applyPaging(Request $request): void
+    {
+        $request->query()->merge([
+            'limit', $this->limit(),
+            'offset', $this->currentOffset(),
+        ]);
```
Sammyjo20: Should we use the $request instance of $pendingRequest?

src/Http/Paginators/OffsetRequestPaginator.php
```diff
@@ -0,0 +1,294 @@
...
+    public function rewind(): void
+    {
+        if (! $this->shouldRewind()) {
+            return;
+        }
+
+        parent::rewind();
+
+        // TODO: Rewind completely, or rewind to originalOffset?
+        $this->offset = 0;
```
Sammyjo20: I would say original offset

src/Http/Paginators/PageRequestPaginator.php
```diff
@@ -0,0 +1,224 @@
...
+    public function rewind(): void
+    {
+        if (! $this->shouldRewind()) {
+            return;
+        }
+
+        parent::rewind();
+
+        // TODO: Rewind completely, or rewind to originalPage?
+        $this->page = 1;
```
Sammyjo20: Original page I reckon

src/Http/Paginators/PageRequestPaginator.php
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /**
+     * @return int|null
+     */
+    public function previousPage(): ?int
+    {
+        return $this->currentPage() > $this->firstPage()
+            ? $this->currentPage() - 1
+            : null;
```
Sammyjo20:
```suggestion
        return $this->currentPage() > $this->firstPage() ? $this->currentPage() - 1 : null;
```

src/Http/Paginators/PageRequestPaginator.php
```diff
@@ -0,0 +1,224 @@
...
+        return $this->lastPage();
+    }
+
+    /**
+     * @return int
+     */
+    public function firstPage(): int
+    {
+        // TODO: Or should we use something from the response?
+        return 1;
```
Sammyjo20: Like with the other paginator, keep it simple - this method is fine

src/Http/Paginators/PageRequestPaginator.php
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /**
+     * @return int|null
+     */
+    public function nextPage(): ?int
+    {
+        return $this->currentPage() < $this->totalPages()
+            ? $this->currentPage() + 1
+            : null;
```
Sammyjo20:
```suggestion
        return $this->currentPage() < $this->totalPages() ? $this->currentPage() + 1 : null;
```

src/Http/Paginators/RequestPaginator.php
```diff
@@ -0,0 +1,281 @@
...
+     * @var bool
+     */
+    protected bool $async = false;
+
+    /**
+     * Whether or not the Paginator will skip rewinding, if it's used in a new loop.
+     * If not, it'll continue where it left off in a previous loop.
+     * While iterators, by design, starts over, Request Paginators should continue.
+     * Otherwise it'll re-send earlier requests.
+     * So make it default not to rewind, unless explicitly told to do so.
```
Sammyjo20: Freakin' love this

src/Http/Paginators/RequestPaginator.php
```diff
@@ -0,0 +1,281 @@
...
+     * @return \Saloon\Contracts\Pool
+     */
+    public function pool(
+        callable|int $concurrency = 5,
+        ?callable $responseHandler = null,
+        ?callable $exceptionHandler = null,
+    ): Pool {
+        // The purpose of a pool is to concurrently send requests.
+        // So 'force' set async.
+        $this->async();
```
Sammyjo20: If the iterator provides request classes, then the pool will use `sendAsync`

Sammyjo20: Or is this just so the paginator is also "async"

src/Http/Paginators/RequestPaginator.php
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    /**
+     * @return ($this->async is true ? \GuzzleHttp\Promise\PromiseInterface : TResponse)
+     */
+    public function current(): Response|PromiseInterface
+    {
+        $this->applyPaging(
+            $request = clone $this->originalRequest,
+        );
```
Sammyjo20: This is clever!

Sammyjo20: Still think modifying the PendingRequest might be better? I can't decide

src/Http/Paginators/RequestPaginator.php
Sammyjo20: isFinished works!

Sammyjo20 (commented):
Love it, it's coming together well

src/Http/Paginators/RequestPaginator.php
juse-less: Yeah, exactly. We mark the internal state to be async, so the paginator will create Promises, instead of requests.
It's a bit of WIP, though, as we still need to figure out how to retrieve total pages when we're doing Promises.

src/Contracts/RequestPaginator.php
juse-less: I re-used the method to check if we're on the last page, or if rewinding is enabled.
So will keep `shouldRewind()` for now.

I added a `rewindingEnabled()` method that explicitly returns the property.
I might rename it to `isRewindingEnabled()`, though.

src/Contracts/RequestPaginator.php
juse-less: Testing `enableRewinding()` as of now. Will see how I feel about it as we go. 😄

src/Http/Paginators/RequestPaginator.php
juse-less: Strictly speaking, we want to send a new request.
Which is why I want to follow the regular sending-flow.

I'm not sure what the difference would be of creating a PendingRequest instead, as we'd still clone the initial Request.

juse-less:
The reason the Laravel team forked opis/closure v3, is because v4 requires a PHP extension.
They wanted to retain backwards compatibility, and not requiring people to have to install the FFI PHP extension.

I only followed the discussion 'as it happened', and haven't kept myself up-to-date.
So I don't know if something around it has changed.

-----
I'll read your feedback a bit more thoroughly, when I get back to this.
Which should be tomorrow, or Tuesday.

I'll push a few updates, with rebase of the 'new' v2 state.
I had issues with my own git fudge-up in the previous PR.
But hopefully they will be solved now, and future PRs shouldn't face the same conflicts and stuff.

One change I did was from your feedback - it'll now automagically rewind if we're on the last page, even if rewinding is disabled.
As part of that I added `rewindingEnabled()`, which returns the property, and used `shouldRewind()` to check if we're on the last page or not.
It made it easier to check `! $this->shouldRewind()`, and bail the `rewind()` method, if not.
Also added `isFirstPage()` and `isLastPage()` as part of that, as well.

juse-less:
For some reason I can't reply to that specific comment.

> I also think we should make sure we can let people customise the query parameters that are set, nothing fancy maybe properties or a method they extend, but I once dealt with an API that used the terminology "top" and "skip" instead of "limit" and "offset" 🤮

I have to work with Jira every day, so I definitely share the feeling for weird APIs.

I thought about it yesterday, but didn't add it that until I saw your comment now.
So you can now find them in this PR.
Each Paginator that defines the 'property' can now handle `with<Property>` to customise the name of the property.
I.e.,

```php
RequestPaginator::withLimitName('maxResults');
RequestPaginator::withLimitName('top');

OffsetRequestPaginator::withOffsetName('startAt');
OffsetRequestPaginator::withOffsetName('skip');

PageRequestPagniator::withPageName('some_other-page.name');
```

Although, I think `using` would be a better terminology here, so we might rename it.

src/Http/Paginators/OffsetRequestPaginator.php
juse-less: Yeah, that's what I'm thinking.
If we were to continue the iteration somewhere, and original offset is not 0/the lowest offset.
_Rewinding_ should be to rewind to the start, imo.
But it could also give unpredictable behaviour, if the first position (original offset/page) is not the lowest, then we'd still send request to the lowest..

src/Http/Paginators/PageRequestPaginator.php
juse-less: Yeah, that's what I'm thinking.
If we were to continue the iteration somewhere, and original offset is not 0/the lowest offset.
_Rewinding_ should be to rewind to the start, imo.
But it could also give unpredictable behaviour, if the first position (original offset/page) is not the lowest, then we'd still send request to the lowest..

src/Contracts/RequestPaginator.php
juse-less: Yeah, that's true.
Since I added naming of the limit property, we can even use that to extract the limit after the first request.

So I'll keep it for now.

src/Http/Paginators/RequestPaginator.php
Sammyjo20: Okay, happy with that now :D

Sammyjo20:
Oh I like that, so you would do something like?

```php
public function resolvePaginator(): Paginator
{
    return PageRequestPaginator::withLimitName('top')->withOffsetName('skip');
}
```

juse-less:
> Oh I like that, so you would do something like?
> 
> ```
> public function resolvePaginator(): Paginator
> {
>     return PageRequestPaginator::withLimitName('top')->withOffsetName('skip');
> }
> ```

Yeah, exactly. That's the plan.
I temporarily removed the Paginator stuff from the Connector, because the TestConnector kept failing (as it needed to define   `resolvePaginator()`).

But once we're done, and we put it back, that's the idea.

Sammyjo20:
YAY!

juse-less:
I'm not quite sure how to solve
```
------ -----------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Line   Http/Paginators/OffsetRequestPaginator.php
------ -----------------------------------------------------------------------------------------------------------------------------------------------------------------------
 14     Class Saloon\Http\Paginators\OffsetRequestPaginator has an uninitialized readonly property $limit. Assign it in the constructor.
        ✏️  Http/Paginators/OffsetRequestPaginator.php
```

I'm fairly certain it's a bug in PHPStan, as we're clearly calling the constructor, which does set the `$limit` property..
I've seen other such issues that PHPStan incorrectly fails on, specifically in the constructor.

juse-less:
@Sammyjo20 Can you have a look at my tests? (https://github.com/Sammyjo20/Saloon/pull/152/commits/c05d7d8e6baf968884894153edb06bd94e9fcd95)
I feel like I've misplaced them, and not quite testing things correctly.

* Sync RequestPaginators seems to work, as I can iterate through all pages successfully in the tests.
* Async RequestPaginators seems to work, judging by the tests.
* Pooled RequestPaginators seems to work, judging by the tests.
* Keep in mind that with concurrent/async requests, we don't know the total requests/pages until the first response is returned.
   Because of this, I've made the RequestPaginator always await until we have a response to grab the relevant data from.
   So the first request (or if we're just lacking a response in memory, for whatever reason) for async and pooled RequestPaginators will in fact be sync.
   You can find that [here](https://github.com/juse-less/Saloon/blob/c05d7d8e6baf968884894153edb06bd94e9fcd95/src/Http/Paginators/RequestPaginator.php#L203-L207).

- We should probably test so all Responses has the proper, sequential page we expect, to ensure the RequestPaginator doesn't skip a Request, or re-do the same, etc.
- I've dropped the RequestPagination serialisation logic for now.
   I'll delete the interface and everything from the branch before we merge.
   It should be possible to add the serialisation bits as a feature PR after v2, without breaking compatibility.
- The continuation feature duplicates the latest page/request every time you continue a loop, because of how PHP iterates Iterators.
- I'm not particularly happy with the code overall.
   * I can't find a proper, abstract, term for paging (page/offset/etc). It's partly why I can't make up my mind or figure out a decent DX.
      Because of that, I've also stripped a lot of stuff from the RequestPaginator interface, because they were duplicating things for offset-based pagination, for instance.
   * It turns out there's quite a few caveats, and more TODOs all over the place (the continuation feature, f.e.).

I'm not really sure how to finish this off.
The lack of proper terminology, DX, and the problems I'm facing is quite frustrating.
So I might need to take a break from the PR, to rethink.

src/Http/Paginators/PageRequestPaginator.php
```diff
@@ -0,0 +1,115 @@
...
+     * @return int
+     */
+    public function totalEntries(): int
+    {
+        // Make sure we have a response.
+        if (is_null($this->currentResponse)) {
+            $this->current();
+        }
+
+        return $this->currentResponse->json('json');
```
Sammyjo20: Are we able to still allow people to overwrite these? We could also provide another variable so people can find the "total" in requests e.g

```php
return $this->currentResponse->json($this->getTotalKey());
```

src/Http/Paginators/PageRequestPaginator.php
```diff
@@ -0,0 +1,115 @@
...
+     * @return int
+     */
+    public function totalPages(): int
+    {
+        // Make sure we have a response.
+        if (is_null($this->currentResponse)) {
+            $this->current();
+        }
+
+        return $this->currentResponse->json('last_page');
```
Sammyjo20: Same with this one, can we have something like:

```php
return $this->currentResponse->json($this->getLastPageKey());
```

Sammyjo20: I feel when people extend this logic they may forget to add:

```php
if (is_null($this->currentResponse)) {
    $this->current();
}
```

So maybe we recommend people extend something like: 

```php
protected function resolveTotalPages();
```

Key-word there, "resolve" - like we have everywhere else in Saloon, then you could do something like:

```php
public function totalPages(): int
{
    // Make sure we have a response.
    if (is_null($this->currentResponse)) {
        $this->current();
    }

    return $this->resolveTotalPages() ?? $this->currentResponse->json('last_page');
}
```

tests/Feature/RequestPaginationTest.php
```diff
@@ -0,0 +1,47 @@
...
+    $responses = [];
+    $superheroes = [];
+
+    foreach ($paginator as $response) {
+        $responses[] = $response;
+        $superheroes = [...$superheroes, ...$response->json('data')];
+    }
+
+    expect($responses)->toHaveCount(4)->each->toBeInstanceOf(Response::class)
+        ->and($superheroes)->toHaveCount(20)->each->toBeArray();
```
Sammyjo20: Since the API is never going to change I would expect to actually see the right superheroes, because this may be doing the same page over and over again and your test wouldn't show it. Let me know if you want the JSON behind the API

tests/Feature/RequestPaginationTest.php
```diff
@@ -0,0 +1,47 @@
...
```
Sammyjo20, line 231: Again, the rest of this test is absolutely fine, but I would be certain what results you are getting back

Sammyjo20: You could also check the pending requests for each response have the right query parameters too, that would be a great test assertion like

```php
expect($responses[0]->getPendingRequest()->query()->all())->toEqual(['page' => 1, 'per_page' => 5]);
expect($responses[1]->getPendingRequest()->query()->all())->toEqual(['page' => 2, 'per_page' => 5]);
expect($responses[2]->getPendingRequest()->query()->all())->toEqual(['page' => 3, 'per_page' => 5]);
expect($responses[3]->getPendingRequest()->query()->all())->toEqual(['page' => 4, 'per_page' => 5]);
```

Sammyjo20: The example is paged but you get what I mean :)

tests/Unit/RequestPaginationTest.php
```diff
@@ -0,0 +1,189 @@
...
+
+namespace Saloon\Tests\Unit;
+
+use GuzzleHttp\Promise\PromiseInterface;
+use Illuminate\Support\LazyCollection;
+use Saloon\Contracts\Response;
+use Saloon\Tests\Fixtures\Connectors\PageRequestPaginatorConnector;
+use Saloon\Tests\Fixtures\Requests\PageGetSuperHeroesRequest;
+
+test('you can yield from a paginator', function (): void {
```
Sammyjo20, line 18: Is this test specifically testing the LazyCollection? If so I would change your test title a little bit

tests/Unit/RequestPaginationTest.php
```diff
@@ -0,0 +1,189 @@
...
+    // TODO: See the test skip() message.
+
+    foreach ($paginator as $response) {
+        $responses[] = $response;
+        $superheroes = [...$superheroes, ...$response->json('data')];
+    }
+
+    expect($responses)->toHaveCount(4)->each->toBeInstanceOf(Response::class)
+        ->and($superheroes)->toHaveCount(20)->each->toBeArray();
+})->skip(<<<'REASON'
```
Sammyjo20: Is this test still needed since you removed the serialization now?

tests/Unit/RequestPaginationTest.php
```diff
@@ -0,0 +1,189 @@
...
+    expect($responses)->toHaveCount(1)->each->toBeInstanceOf(Response::class)
+        ->and($superheroes)->toHaveCount(5)->each->toBeArray();
+
+    foreach ($paginator as $response) {
+        $responses[] = $response;
+        $superheroes = [...$superheroes, ...$response->json('data')];
+    }
+
+    expect($responses)->toHaveCount(5)->each->toBeInstanceOf(Response::class)
+        ->and($superheroes)->toHaveCount(25)->each->toBeArray();
```
Sammyjo20, line 215: Again I would have a pest function that is a shortcut to get the full array of superheroes in the right order so you know exactly the API has returned with what you expect

Sammyjo20 (commented):
Alright overall great tests, I would use the superheroes.json file and decode it in a global pest helper function so when you compare the $superheroes array you can actually check if yours equals the exact array and order that you expect.

In terms of DX I'm reluctant to have the `HasPagedPagination` trait as it pollutes the public API a little bit with methods like "usePageName" etc.

Instead, how do you feel about what I said before where we can make people create their own paginators that extend our base one? Let me explain

We can have three classes
- Abstract Paginator class, like what you have now that works with both paged + limit/offset

Then these: 
- OffsetPaginator class, 
- PagedPaginator class 
> Not abstract but define defaults like you have already in your traits

Next people could try using those paginators directly, but if they have any kind of custom implementation, we just ask, to extend the paginator class and then modify the methods. You could implement a "setup" method that is executed so people could do something like this:

```php
class CustomPaginator extends OffsetPaginator
{
     public function setUp(): void
     {
          $this->usingLimitName('top');
          $this->usingOffsetName('skip');
     }
}
```

The `setUp` method could also be called something like `configure` to make it more developer friendly.

Alternatively, you could move to protected functions that could be extended, this is very saloon like - similar to how you define the base URL, endpoint etc

```php
class CustomPaginator extends OffsetPaginator
{
    protected function resolveLimitName(): string
     {
          return 'top';
     }

     protected function resolveOffsetName(): string
     {
          return 'skip';
     }

     protected function resolveResultsProperty(): string
     {
          return 'data';
     }
}
```

What are your thoughts? We could even implement default paginators like LaravelPaginator.

Sammyjo20:
@juse-less are Twitter DMs broken for you?

juse-less:
* ~~The continuation feature duplicates the latest page/request every time you continue a loop, because of how PHP iterates Iterators.~~

This turned out to be because of how *I* handled the 'failsafe' logic in the RequestPaginator implementations (Paged + Offset).
Brainfarted that one..
I've added comments around the relevant, sensitive code in the base RequestPaginator, that caused it to send extra requests.

So I think all features for v2 release should now _work_.
Just gotta solve DX, testing, and the feedback @Sammyjo20 just gave, as I wrote this.

Sammyjo20:
We also need to consider that the JSON:API specification doesn't actually say anything about paginated resources having to have a "total". Do we have the ability to extend the "isLastRequest" logic? Sorry I probably skimmed past it.
 
Screenshot 2023-02-08 at 21 57 29

juse-less:
@Sammyjo20 Yep - completely broke down after our excitement when I figured out my brainfart with the continuation feature. 😄

Sammyjo20:
Twitter couldn't contain our excitement.

juse-less:
Regarding the JSON API spec - that one is a bit tricky.
The base `RequestPaginator` is checking `isFinished()`, which is implemented by classes extending it.
Currently we do check for total entries (offset) and last_page (page) keys to determine total amount of requests possible for the resource vs the current page.
They can be extended, however, to override that (or extend the `RequestPaginator` itself, or even a completely custom one).

Would it make sense to add a 3rd paginator that iterates while there are pages?
The problem with that is pooling and async requests - if we don't know the total amount of pages/requests, we don't know the next page, and which one is the last page.

tests/Unit/RequestPaginationTest.php
juse-less: That test was written for the continuation feature, which is used to break out of loops, and then keep iterating in new ones.
I'm not sure if it'd be used without serialisation, though.
I'll leave that call to you.

Sammyjo20:
@juse-less My original idea was to have a method like `isLastPage($response) and you could use the current response instance to determine if you're on the last page, which means you don't need to know the total amount pages if you get the developer to define if it's on the last page. It could be anything like:

return empty($response->json())

or 

return is_null($response->json('next_page'))

but the default implementation is what you already have.

Sammyjo20:
I'm happy to try and implement it on your branch if you want me to? But I don't want to step on your toes :D

juse-less:
The problem I see with this is with concurrent requests (like the Pool).
In order to check if we're on the last page, we'd need to await the Response before sending a new Request.
Which would make the whole process synchronous.

I'm not sure if there are other ways to check how many requests are possible so you can keep sending without requesting non-existing pages, or so.

> I'm happy to try and implement it on your branch if you want me to? But I don't want to step on your toes :D

Appreciate it. I'd happily hand over my branch for a while. ;D
Give it a try and see if you can come up with a solution. It's always good with a new set of eyes.

Sammyjo20:
Hmm that's true, I see what you mean around concurrency and pools ☹️ you're right because it's concurrent, it needs a constant buffer of requests, and in order to do that we need to know how many requests to make. 

> I'd happily hand over my branch for a while

I'm a little scared, I don't want to break anything 😆 

So we could switch to class-based paginators for now, but we still have the problem around what if there is no "total" displayed. 

Hmm what do you think of the class-based paginators? I'm starting to think the traits could equally work too. Just trying to think of the nicest ways for a developer. Being a trait is one less class to make... but equally another class isn't a bad thing especially with the customisation plus we can use protected methods and not public ones

tests/Unit/RequestPaginationTest.php
Sammyjo20: Oh okay, I'd say keep it then

juse-less:
We could always make the properties for the keys public.
I think that'd be the easiest option.

I loved your idea around a public `configure()` method. At first I thought it could accept a callable, which received the paginator.
But.. that'd be fairly pointless as you could just call the properties/methods without it.

But yeah.. public properties so that people don't have to extend an existing paginator just to change key names, or a protected `configure()` method (through the base paginator) that people can extend/override to set the properties.

But.. I mean..
```php
// Connector

public function paginate(Request $request, mixed ...$additionalArguments): RequestPaginatorContract
{
    return new class ($request, ...$additionalArguments) extends RequestPaginator
    {
        protected function configure(): void
        {
            $this->limitKey = 'all_the_things';
            $this->offsetKey = 'head_start';
        }
    };
}
```

> I'm a little scared, I don't want to break anything 😆
Don't worry - I need to rebase the whole thing from the start.
1. I accidentally got your earlier history in it.
1. I've gone back and forth between method names, and need to clean up the history anyway.

----------
Oh right!
Twitter released that increased tweet size earlier today. I guess it couldn't handle the load. ^^

Sammyjo20:
What about...

```php
public function paginate(Request $request, mixed ...$additionalArguments): RequestPaginatorContract
{
   return  PagedPaginator::make($request, ...$additionalArguments)->setLimitKey('top')->setOffsetKey('limit');
}
```

That'll cover 90% of the use case, but if people need to customise the methods, they could extend it and return their own?

Also I think in keeping it consistent, we should have an internal `paginate` method and then users can extend a `resolvePaginator` protected method when they add the `HasPagination` trait.

```php
class MyRequest extends Request
{
      use HasPagination;
      
      protected function resolvePaginator(): Paginator
      {
            return  PagedPaginator::make()
                      ->setLimitKey('top')
                      ->setOffsetKey('limit');
      }
}

$request = new MyRequest;
$request->paginate()
```

The resolve paginator method doesn't need the additional arguments or requests because you would still pass the request/additional arguments when you call ->paginate()

juse-less:
That would take us back to where we started, wouldn't it?
I only exported the name/key methods to reusable traits because they are common.
Because the offset paginator doesn't have pages, I don't have the page-related methods on the interface, and vice versa.
They are only added on requests or paginators by using the traits.

Regarding a protected `resolvePaginator()` method - I originally had that, but saw no reason as our public `paginate()` method was just calling that one anyway.
Good point about consistency, though - I can add it back.
Or.. rather.. I can add the whole trait back. 😛 

Currently, the base `RequestPaginator` takes a `Connector` and a `Request` as constructor arguments.
The `Request` is used as a template for additional requests, and is readonly to prevent being changed (or at least replaced in the property).

Sammyjo20:
> Or.. rather.. I can add the whole trait back. 😛

Ignore me, this is just going to be on the connector. 

> That would take us back to where we started, wouldn't it?

I don't think so, the PagedPaginator has a page query parameter and the OffsetPaginator has an offset/limit query parameter, so people could configure it inside of the `resolvePaginator` method on the _connector_   

> I only exported the name/key methods to reusable traits because they are common.

But if the configurable methods are on the paginator class itself, that means no extra methods on requests/connectors as it all stays within the paginator's public APIs. 

The following can also be serialized too, as there are no closures

```php
protected function resolvePaginator(): Paginator
{
      return  PagedPaginator::make()
                ->setLimitKey('top')
                ->setOffsetKey('limit');
}
```

Sammyjo20:
Ignore what i said about traits, my proposal is to add the `resolvePaginator` method and use public properties on the paginator classes to configure said paginator inside of the `resolvePaginator` method. No traits needed, it will all be supported as a first-party feature

This is how the paginate method could look on the connector:

```php
public function paginate($request): Paginator
{
      return $request->resolvePaginator($request) ?? $this->resolvePaginator($request);
}
```

Then actually I guess `resolvePaginator` needs to be public on the request? 

Sorry Juse, I'm trying to figure this one out - I'm a bit tired tonight haha. I feel like we're getting there thouhg

juse-less:
> Sorry Juse, I'm trying to figure this one out - I'm a bit tired tonight haha. I feel like we're getting there thouhg

Haha. No worries. I think we both need a break. It feels like we're both kinda talking about the same thing, but still end up misunderstanding each other. 😅 
I've rewritten the below message more times than I can count on one hand, but'll keep the latest revision.
I'll have a look at all your feedback today/tomorrow (depending on timezones), or on Friday.
Like you said - getting there, though!
Feature wise, all issues are resolved from, what I can tell.
Just need to polish things make it tested, and provide a proper public API.

----------
Sounds like a plan!
I'll make the key properties public.
~~Want me to add a HasPaginator trait with the our default `paginate()` method, as well as the `resolvePaginator()` method, or just define them directly on the default Connector?~~

~~Yeah, but that's how it works right now.
Because the _requests_ are what carries the query parameters, I opted to use the trait to define them, so others can use it on their custom requests for convenience (but not required).
But can also be put on the paginators (as they forward query parameters).~~

~~Currently, those traits are also only used on the paginators, and works the way you describe.
I'm guessing it's a bit confusing because of the fact that I put them under the Request trait namespace, as opposed to the RequestPaginator?
My thought-process was that the query parameters are request-related, and can be used by both (but we only use them on our pre-defined paginators).~~

Sammyjo20:
I agree. Sorry I don’t mean to keep getting confused either, I think this whole doing it over message can be tricky 😅

I am pretty exhausted actually, so I’m going to have an early-ish night. We’ll crack this though! When I’m not tired I’ll try to do some coding on it. I feel we’re so close.

Sammyjo20:
Re-reading your crossed out message that does make sense. Geez I really must be tired as it’s not gone in at all 😅

I trust your judgement, I might actually load this up in a project and take it for a test drive before I try to unpick it further

juse-less:
Speak to you tomorrow then. Hopefully Twitter is back up, so we can take discussions back-and-forth without spamming the PR. 🙂

Like I also said before - you're free to use my branch, etc.
It needs rebasing anyway, as I somehow got the history of an entire multiverse in this PR. 😂

Sammyjo20:
> Would it make sense to add a 3rd paginator that iterates while there are pages?

This could be a good option :) or maybe define a “noPage” option either one of the paginators configure method like 

$paginator->withoutTotal()

that will make it throw an exception if you attempt to use “pool” because it cannot be done. Let’s not try to spend too much time on that, if there is no count, we can’t use pools. 

Then people can extend the isFinished method.

Sammyjo20:
I'm having a play and so far it works beautifully - I'm using the superheroes API in a Laravel app. I'm really glad I tested this now because I actually was completely wrong about the traits! I thought "HasPagedPagination" and "HasOffsetPagination" actually lived on the request/connector but they're just internal traits that you use!

I was getting really confused because I thought you were adding them to the request, so that makes more sense. I may have more suggestions on the code as I understand it a bit more but it's definitely super easy to setup. I decided to just use the `HasRequestPagination` interface and just define the paginator. It was certainly easier than using the resolvePaginator as it's just one less step. **Although I'm still undecided on whether to use resolve* because that would standardise things**

When I tried using `->collect()` I was getting this error:

Screenshot 2023-02-09 at 22 59 09

Sammyjo20:
I think it's because we're now using the `$this->json()` method, but using a closure and `yield from` works.

Screenshot 2023-02-09 at 23 00 29

Sammyjo20:
Okay so a few things I've noted about the collect function.


- When using the collect method, you lose access to the original response since it uses `json` under the hood. Perhaps we should add a new property like `asResponses` which by default is false, but if people want the raw response, they can set it to true?

- If you don't use any methods on the collection, it will come out as a collection of 4 arrays, each containing the 5 superheroes. I used the `collapse` method which basically combines them into one big array of 20 superhero arrays. I'm wondering if we should add an argument to the `collect` method which is "collapse: true" and so by default it will automatically smash all the items in the results together, like it's one big array.

I might make a small commit to change this slightly.

Edit - however I absolutely LOVE it @juse-less you've done a fantastic job, and I really feel the DX is close. I agree with you it needs a little tidy up internally, but we can get this close. I

Sammyjo20:
We can do this too! YAY! This was what I was talking about. You already implemented it lol I was super tired last night

Screenshot 2023-02-09 at 23 17 03

Sammyjo20:
If it's okay with you (which I'm sure it is as I asked before) I might do some commits on Saturday going through a few things like:

- Removing the `HasRequestPagination` trait
- Renaming the interface to `HasPagination`
- Removing the `...$additionalArguments`
- Fixing the collect method and adding additional functionality like collapsing results and returning responses 
- Tidying up the classes (if needed)

Let me know if you don't want me to, but at least them I feel you can have a rest and all I'm doing is just giving it that final polish up.

juse-less:
Regarding the broken `collect()` method:
~~It's actually a bug in Laravels `LazyCollection`, per https://github.com/laravel/framework/blob/e92cbda8cb6bc9b416ed89c2339722d462111269/src/Illuminate/Collections/LazyCollection.php#L40.
It takes an iterable, and should therefore work.~~
While still technically true, as a Generator is an iterable, I realised the above isn't quite right.
I read my actual code, and realise I'm giving it a Generator and not the array I meant was iterable.
So I'll sort that this lunch, as well, as it's easily fixable.

----------
Regarding the purpose of the `collect()` method:
It's purpose is to collect the data from Responses.
If you need to iterate Responses, you can just iterate the paginator, or, possibly `iterator_to_array($paginator)`, to turn them into individual arrays.
Although I guess it does make sense in this context to have the `collect()` method collect the Responses by default.
Then add the additional parameter, like you mentioned, to make it collect each data instead.

----------
Regarding the `use<Key>Name()` methods:
They are weird to me, and, if we're to keep them - rename them to `set<Key>Name()`.
Not sure why I'm so allergic to using setters these days? Damaged by Laravel? 😄

Again, the reason they are in the trait is so that others _can_ add them to their requests.

Like..
```php
class GetMyItemsRequest extends Request
{
    use HasPagedPagination;
}

$request = GetMyItemsRequest::make()->setPageName('other_name')->setPage(4);
```

So, as you can see, it's a convenience trait to optionally add the same methods, so _others_ can semantically set pages instead of `$request->query()->add('page', 4)`.
However.. I realised now that when I added the `use<Key>Name()` methods to them, it became a tad weird from a Request perspective..

----------
I'll try sort some stuff at my lunch today - but then you're perfectly fine to take over.
There are plenty of tests that needs adding and fixing, as well.

----------
I'd also like to take the opportunity to thank you for giving me this opportunity.
I've had a hard time getting into open source, but this has been learning from many perspectives.

----------
Edit: Corrected the first answer.
Edit 2: Added another answer/reflection around the pagination traits.

Sammyjo20:
In a DM you mentioned that the traits could be removed in favour of adding the methods directly to each paginator, I think I’ll do that too, since the traits aren’t being used in multiple places and it’s not really needed

Sammyjo20:
Todo Sam: we also discussed adding a pre-implemented protected “configure” method on the paginator abstract class that can be easily extended. This is useful for people who want to extend the paginator but not overwrite the constructor to configure the paginator.

Sammyjo20:
I've made a bunch of changes to the paginator, just a note to self - I need to write some more tests to improve code coverage but I think we're almost there now (assuming you don't find anything that you hate 😉)

juse-less:
@Sammyjo20 I've added you as collaborator on my fork now.

I think the reason you can't push is that you created the PR, and wasn't able to give permission to push to my PR branch.
Normally that option is pre-ticked when I create PRs, but that option isn't available here, and I can't edit the PR as you created it.

But you should be able to push to my branch now.

tests/Pest.php
```diff
@@ -66,3 +66,8 @@ function paginationMockClient(string $prefix): MockClient
         }
     ]);
 }
+
+function superheroes(): array
```
juse-less, line 80: While it is a method to grab test data, I think it'd be nice to add array shape, as it could be used in tests for autocompletion.

```suggestion
/**
 * @returns array{
 *     id: int<1, max>,
 *     superhero: string,
 *     publisher: string,
 *     alter_ego: string,
 *     first_appearance: string,
 *     characters: string,
 * }
 */
function superheroes(): array
```

Sammyjo20 (commented):
Self review

Sammyjo20 (commented):
Self review

Sammyjo20:
@juse-less There's a couple of PHPStan issues I understand but not sure about how to fix...

juse-less:
Most of them already existed before, and has been bugging me.

```
Error: Method Saloon\Http\Paginators\CursorPaginator::getCursor() should return int|string|null but returns array|string|null.
```
<details>
<summary>⬆️ Details</summary>

I think this one has to be explained a bit, so let's break it out:
`$query` is an array, as it's constructed by [`parse_str()`](https://www.php.net/manual/en/function.parse-str.php).
`parse_str()` takes a string of URL query params, and turns them into an array.

So far so good.. but.. if you think about..

Query param values can only really be strings (not int, float, etc), right?
So each element, if existing, would be a string.
But.. [you can also have query params that are actually arrays (`param1[]=arg1&param2[]=arg2`)](https://www.php.net/manual/en/function.parse-str.php#refsect1-function.parse-str-examples).
So from a `parse_str()` perspective, the type hint would be `array<array<string>|string>` or `array<string[]|string>`.
But since we're trying to return a specific element, that could very well be a flat string, or one of those query params that has multiple values.
That makes it `array<string>|string` or `string[]|string`.
However, there's a possibility the key for the element doesn't exist, so we return null ourselves.

So, to summarise, I believe the return type should really be `array<string>|string|null`, to be more correct.

</details>

```
Error: Call to an undefined method GuzzleHttp\Promise\PromiseInterface|Saloon\Contracts\Response::json().
```
<details>
<summary>⬆️ Details</summary>

This is because of the dispatcher that I DMd you about yesterday.
Technically somewhat of a false-positive, but can in theory happen if someone was to extend/replace the logic.
The reason is that we can't correctly type hint `Dispatcher::execute()`s return.
We _know_ it'll return a PromiseInterface when PendingRequest->isAsync() is true. And a Response when false.
But PHPStan doesn't understand that.
There are newer assertion tags, but I wasn't able to use them to sort the error when I tried a bit.
The only idea I can come up with is if we remove the Dispatcher, that we talked about, but have the PendingRequest send itself through the Connector.
It'd solve that error, but retain the logic for the sendAndRetry method we struggled with.

Worst case.. add a bool, `$async`, to the execute() method, to be used for that explicitly......
I'd advice against it, though.
</details>

```
Error: Parameter #1 $requests of method Saloon\Contracts\Connector::pool() expects (callable(Saloon\Contracts\Connector): iterable<(int|string), GuzzleHttp\Promise\PromiseInterface|Saloon\Contracts\Request>)|iterable<(int|string), GuzzleHttp\Promise\PromiseInterface|Saloon\Contracts\Request>, $this(Saloon\Http\Paginators\Paginator) given.
```
<details>
<summary>⬆️ Details</summary>

I'm not entirely sure why it's complaining.
The problem is that I've type-hinted the iterables..
So if you remove the diamonds at both iterables (`iterable<array-key, ...>` -> `iterable`), it'll resolve (but then have incomplete types).
</details>

Sammyjo20:
Thank you for your help on the generics @juse-less I've tried updating them and I can't quite seem to get it to work tonight, I will try again next time I'm coding. When I removed the detail on the last one it added new errors

Screenshot 2023-02-14 at 22 08 19

juse-less:
I tested a bit, and it looks like what I originally had is actually correct.
I removed the iterable stuff myself because PhpStorm was complaining. But it looks like a bug in PhpStorm, now that I tested a bit more.

<details>
<summary>See my recording of a quick test</summary>


https://user-images.githubusercontent.com/76173223/218908816-3b15f4e7-f6af-4bf3-b842-58e307b86e0a.mp4
</details>

So to summarise - for the `$requests`, including Pool, Connector, etc, the type hint should be
```php
* @param iterable<\GuzzleHttp\Promise\PromiseInterface|\Saloon\Contracts\Request>|callable(\Saloon\Contracts\Connector): iterable<\GuzzleHttp\Promise\PromiseInterface|\Saloon\Contracts\Request> $requests
```

I guess it makes sense to specify its keys, as well. Kinda redundant, but sometimes PHPStan wants them when giving iterables to methods or classes using templates.
So that'd leave us at.
```php
* @param iterable<array-key, \GuzzleHttp\Promise\PromiseInterface|\Saloon\Contracts\Request>|callable(\Saloon\Contracts\Connector): iterable<array-key, \GuzzleHttp\Promise\PromiseInterface|\Saloon\Contracts\Request> $requests

```

Sammyjo20 (commented):
Self review

phpstan.baseline.neon
```diff
@@ -96,6 +96,21 @@ parameters:
 			path: src/Repositories/Body/MultipartBodyRepository.php
 
 		-
-			message: "#^Parameter \\#2 \\$ of callable callable\\(\\$this\\(Saloon\\\\Repositories\\\\Body\\\\StringBodyRepository\\), TValue\\)\\: void expects TValue, array\\|float\\|int\\|string\\|false\\|null given\\.$#"
+			message: "#^Parameter \\#2 \\$ of callable callable\\(\\$this\\(Saloon\\\\Repositories\\\\Body\\\\StringBodyRepository\\), TValue\\)\\: void expects TValue, array\\|float\\|int\\|string\\|false\\|null given\\.$#" z
 			count: 2
 			path: src/Repositories/Body/StringBodyRepository.php
+
+    -
```
juse-less: Spaces to tabs
```suggestion
		-
			message: "#^Call to an undefined method GuzzleHttp\\\\Promise\\\\PromiseInterface\\|Saloon\\\\Contracts\\\\Response\\:\\:json\\(\\)\\.$#"
			count: 1
			path: src/Http/Paginators/Paginator.php
```

phpstan.baseline.neon
```diff
@@ -96,6 +96,21 @@ parameters:
 			path: src/Repositories/Body/MultipartBodyRepository.php
 
 		-
-			message: "#^Parameter \\#2 \\$ of callable callable\\(\\$this\\(Saloon\\\\Repositories\\\\Body\\\\StringBodyRepository\\), TValue\\)\\: void expects TValue, array\\|float\\|int\\|string\\|false\\|null given\\.$#"
+			message: "#^Parameter \\#2 \\$ of callable callable\\(\\$this\\(Saloon\\\\Repositories\\\\Body\\\\StringBodyRepository\\), TValue\\)\\: void expects TValue, array\\|float\\|int\\|string\\|false\\|null given\\.$#" z
```
juse-less: Remove trailing  `z`.
```suggestion
			message: "#^Parameter \\#2 \\$ of callable callable\\(\\$this\\(Saloon\\\\Repositories\\\\Body\\\\StringBodyRepository\\), TValue\\)\\: void expects TValue, array\\|float\\|int\\|string\\|false\\|null given\\.$#"
```

Sammyjo20 (commented):
Self review

Sammyjo20:
We're finally good to go! Woohooo! Thanks @juse-less for the help with this <3

PR #153, opened 2023-02-05
Fix usages of `expectDeprecationMessage`

Sammyjo20:
This PR fixes issues with tests failing with the latest version of PHPUnit


Sammyjo20 (commented):
Self review

PR #155, opened 2023-02-06
Fix | V1 Encoded Files In Fixtures

Sammyjo20:
There was an issue which was raised in discussion #148 where if a request returned a downloadable file, the fixture could not be recorded because it threw an exception "Malformed UTF-8 characters, possibly incorrectly encoded"

@derheyne Came up with a brilliant idea to `base64_encode()` the response body if it cannot detect UTF-8 encoding. This is also backwards compatible because of the added `encoding` option.


Sammyjo20 (commented):
Self review

Sammyjo20:
@derheyne Let me know if you're happy with this and I'll roll it out

derheyne:
Looks good to me! Thank you so much for the fast implementation :)

Before you merge it let me quickly test it against our code base.

derheyne:
Works perfectly in our code base!

Sammyjo20:
Great thank you for testing!

PR #159, opened 2023-02-12
Fix links and update alt text

tbarn:
Hey! I was checking out Saloon for something I am hacking around on and noticed some of the links were broken in the readme, so I gave it a tiny fix. I wasn't sure if I should PR it to v1 or v2.


Sammyjo20 (approved):
Thank you so much! 🤠

Issue #154, opened 2023-02-05
Ratelimit handling & Pagination


anthonyvancauwenberghe:
Hi,

For a project these past few months I have been using a simpler, self created implementation of saloon (wish I had known this amazing package existed sooner). I've been looking through the documentation & code (of both v1 & v2), but i'm missing 2 (common) features that I need quite often while working with external api's: 
the first one is pagination handling (for which I see there is already a draft? https://github.com/Sammyjo20/Saloon/pull/152). 
The second is rate limit handling.

Are you open to accepting PR's for rate limit handling? Since I'm new to this package could you give some basic pointers on how/where to start with this.  

Kind regards,
Anthony

Sammyjo20:
Hey @anthonyvancauwenberghe 

Many thanks for your message and your kind words on Saloon, please excuse the missing pages in the v2 docs, I'm still working on them in a flurry!

Ah! Well yes pagination is in progress, @juse-less and I are working on it together for version two. 

Pagination & rate limiting where my two biggest bug bears with APIs too. I've implemented rate limiting with Laravel's jobs before but I'd LOVE to discuss how we could implement it with Saloon. Do you have any ideas around it? It'll be good to have a method we agree on before a PR is made

Many thanks!

juse-less:
@Sammyjo20 Off the top of my head, to stay agnostic, we could possibly use [PSR-6](https://www.php-fig.org/psr/psr-6/), or [PSR-16](https://www.php-fig.org/psr/psr-16/) to store the throttling.
These are caches, but are defined as PSRs, making it more versatile/agnostic than implementing it for specific solutions (like Laravel), or an internal storage, or so.
Then people can provide their PSR-supported caching, like something from Laravel, Symfony, or so.

In terms of implementation, at least external APIs should return with HTTP 429 if you're being throttled.
These also contains datetime/interval for when you can try again.
This datetime/interval usually when the throttle is lifted.
It's also used (in Laravel, for instance) on maintenance mode, at which point it's more of 'try again later' datetime/interval.
But for actual throttling, it should be static, as they'd know when your throttling lifts.

Saloon also already has an Exception for HTTP 429, which might be useful for that.

anthonyvancauwenberghe:
Hi @Sammyjo20,

Happy to hear you're open to PR's!

So in my head the rate limit interface could roughly look something like this:

isThrottled() -> e.g. $response->statusCode === 429
retryAt() -> extract from response (or cache) 
[...]

PR #160, opened 2023-02-13
Request delays

anthonyvancauwenberghe:
Hi Sammy,

Part of the solution to implementing ratelimiting (#154).

Let me know what you think.


src/Contracts/HasDelay.php
```diff
@@ -0,0 +1,17 @@
...
+use Saloon\Repositories\Body\IntBodyRepository;
+
+interface HasDelay
+{
+    /**
+     * Access the delay
+     *
+     * @return IntBodyRepository
+     */
+    public function delay(): IntBodyRepository;
```
Sammyjo20: At first, I was initially against having a `IntBodyRepository` but it reduces the number of methods, e.g:
`$request->delay()->set()` and `$request->delay()->get()` is nicer than having `withDelay` and `getDelay` 👍 Nice idea you had here.

src/Contracts/Request.php
```diff
@@ -6,7 +6,7 @@
 
 use Saloon\Enums\Method;
 
-interface Request extends Authenticatable, CanThrowRequestExceptions, Conditionable, HasConfig, HasHeaders, HasMiddlewarePipeline, HasMockClient, HasQueryParams, Makeable
+interface Request extends Authenticatable, CanThrowRequestExceptions, Conditionable, HasConfig, HasHeaders, HasMiddlewarePipeline, HasMockClient, HasQueryParams, HasDelay, Makeable
```
Sammyjo20, line 9: Please can we also have the `HasDelay` interface/trait on the connector too? So people can define a global delay if they need to

src/Http/PendingRequest.php
```diff
@@ -193,6 +193,9 @@ protected function mergeRequestProperties(): static
             ->merge($connector->middleware())
             ->merge($request->middleware());
 
+        $delay = $request->delay()->all();
+        $this->delay()->set($delay);
```
Sammyjo20: When we add connector delays too, do we need to have `merge` method on your `IntBodyRepository`?

Sammyjo20: See my other body repository classes on how I do the merging 👍

src/Http/Senders/GuzzleSender.php
```diff
@@ -172,6 +172,10 @@ protected function createRequestOptions(PendingRequest $pendingRequest): array
             $requestOptions[$configVariable] = $value;
         }
 
+        if ($pendingRequest->delay()->all() > 0) {
```
Sammyjo20: Could you use `isNotEmpty()` here instead? This follows the convention of the body repositories?

src/Http/Senders/SimulatedSender.php
```diff
@@ -70,6 +70,10 @@ public function sendRequest(PendingRequest $pendingRequest, bool $asynchronous =
 
         // We'll return the synchronous response directly
 
+        if($pendingRequest->getRequest()->delay()->all() > 0) {
```
Sammyjo20: We should be checking the delay on the PendingRequest directly instead of getting the request through the PendingRequest `$pendingRequest->delay()->all()`

Same as my comment above, can we do `$pendingRequest->delay()->isNotEmpty()` ?

src/Traits/RequestProperties/HasDelay.php
```diff
@@ -0,0 +1,32 @@
...
+trait HasDelay
+{
+    /**
+     * Request Delay
+     *
+     * @var IntBodyRepository
+     */
+    protected IntBodyRepository $delay;
+
+    public function delay(): IntBodyRepository
```
Sammyjo20: Add missing docblock here please

src/Repositories/Body/IntBodyRepository.php
```diff
@@ -0,0 +1,84 @@
...
+
+declare(strict_types=1);
+
+namespace Saloon\Repositories\Body;
+
+use Saloon\Exceptions\UnableToCastToStringException;
+use Saloon\Traits\Conditionable;
+use Saloon\Contracts\Body\BodyRepository;
+
+class IntBodyRepository implements BodyRepository
```
Sammyjo20: Could we rename this to `IntegerBodyRepository`? Just personal preference :)

tests/Unit/DelayRequestTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+    $mockClient = new MockClient([
+        MockResponse::make(['name' => 'Sam']),
+    ]);
+
+    $start = microtime(true);
+    connector()->send($request,$mockClient);
+    expect(microtime(true) - $start)->toBeGreaterThan(1);
+});
+
```
Sammyjo20: Please could we also write some additional tests based on my feedback above?

- Testing the delay works when on a connector
- Testing the delay works when both connector and request (should we merge or make the request's delay take priority?)
- Unit tests for the IntegerBodyRepository, making sure the methods all work as expected (I can do this if you want me to)

I also noticed these tests use mocking which means we're not testing the Guzzle implementation ,please can we also write a sync/async test that sends a real request with Guzzle?

Sammyjo20 (changes requested):
Many thanks for this awesome PR @anthonyvancauwenberghe! I hope you don't mind my comments.

src/Contracts/HasDelay.php
anthonyvancauwenberghe: I actually implemented it with get/set Delay first. But i didn't like it so switched to how you implemented it! :)

src/Contracts/Request.php
anthonyvancauwenberghe: Yes i'll add it

src/Http/PendingRequest.php
anthonyvancauwenberghe: I'll check it

src/Http/Senders/GuzzleSender.php
anthonyvancauwenberghe: I did this on purpose to avoid sleep exceptions from negative integers. Now we check if the delay is >0 and if not there is no need for delaying the request.

src/Http/Senders/SimulatedSender.php
anthonyvancauwenberghe: Will add delay method on pending request.

src/Repositories/Body/IntBodyRepository.php
anthonyvancauwenberghe: sure!

tests/Unit/DelayRequestTest.php
anthonyvancauwenberghe: - I'll test delays on connectors aswell
- i'll add a test that checks both (request takes priority indeed).
- I'll add unit tests for intrepo

No i tested both the simulation sender & guzzlesender.

src/Http/PendingRequest.php
anthonyvancauwenberghe: I didn't add a merge method on Integerbodyrepository, I didn't really see the point as it would just replace the value.

src/Http/PendingRequest.php
```diff
@@ -232,6 +236,15 @@ protected function mergeBody(): static
         return $this;
     }
 
+    protected function mergeDelay(): static
+    {
+        $this->request->delay()->isNotEmpty() ?
```
anthonyvancauwenberghe, line 238: Request delay takes priority over connector

src/Http/PendingRequest.php
Sammyjo20: I can see your `mergeDelay` method above you added so this can be removed.

src/Http/Senders/SimulatedSender.php
```diff
@@ -70,6 +70,11 @@ public function sendRequest(PendingRequest $pendingRequest, bool $asynchronous =
 
         // We'll return the synchronous response directly
 
+        if($pendingRequest->delay()->greaterThan(0)) {
+            $sleep = ($pendingRequest->delay()->all() *1000);
+            usleep($pendingRequest->delay()->all() *1000);
```
Sammyjo20:
```suggestion
            usleep($pendingRequest->delay()->all() *1000);
```

src/Repositories/Body/IntegerBodyRepository.php
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * Set a value inside the repository
+     *
+     * @param int|null $value
+     * @return $this
+     */
+    public function set(mixed $value): static
+    {
+        if(! is_int($value) && ! is_null($value))
+            throw new \InvalidArgumentException('The value must be an integer or null.');
```
Sammyjo20:
```suggestion
        if(! is_int($value) && ! is_null($value)) {
            throw new \InvalidArgumentException('The value must be an integer or null.');
        }    
```

src/Repositories/Body/IntegerBodyRepository.php
```diff
@@ -0,0 +1,112 @@
...
+
+    public function lesserOrEqualThan(int $value): bool
+    {
+        return $this->data <= $value;
+    }
+
+    public function equalTo(int $value): bool
+    {
+        return $this->data === $value;
+    }
```
Sammyjo20, line 102: Please add docblocks for these methods :)

Sammyjo20 (changes requested):
Few other changes, with regards to the Guzzle sender, we still need to write tests for it

Note - when you use a MockClient, the GuzzleSender is swapped out for the SimulatedSender

anthonyvancauwenberghe:
I use both in my tests. One with mock and one without. Both get tested.

Sammyjo20 (approved):
Sorry I've re-read your tests and I saw you were testing both. Need more coffee! ☕️

This is a great PR, so thank you for sorting this. I will merge this later today.

src/Http/PendingRequest.php
```diff
@@ -193,6 +194,9 @@ protected function mergeRequestProperties(): static
             ->merge($connector->middleware())
             ->merge($request->middleware());
 
+        $delay = $request->delay()->all();
+        $this->delay()->set($delay);
```
Sammyjo20: @anthonyvancauwenberghe Do you need these two lines:

```php
$delay = $request->delay()->all();
$this->delay()->set($delay);
```
Because you have `mergeDelay` above?

Sammyjo20 (changes requested):

anthonyvancauwenberghe:
Ready to merge

Sammyjo20 (approved):

Sammyjo20:
Great work on this @anthonyvancauwenberghe. Thank you for contributing.

PR #165, opened 2023-02-19
Draft: Feature | V2 Debugging

juse-less:
Todo:

- [ ] Add Telescope + Laravel Log drivers to the laravel package


juse-less:
@Sammyjo20 Go wild - you're free to edit anything in my PoC.
Let me know if you're unable to edit the PR post, and I'll fill it in.

I opted to use `syslog()`, and `error_log()` directly to not pull in additional dependencies.
Something like Monolog has a lot more sources, though (including syslog and error log).
Not sure it's worth it from a debugging perspective, though.

src/Debugging/DebugData.php
```diff
@@ -0,0 +1,70 @@
...
+    public function response(): ?Response
+    {
+        return $this->response;
+    }
+
+    public function statusCode(): ?int
+    {
+        return $this->response?->status();
+    }
+}
```
Sammyjo20, line 128: Rename these getters to prefix with `get` like `getUrl`, `getMethod` etc.

src/Debugging/Drivers/ErrorLogDebugger.php
```diff
@@ -0,0 +1,55 @@
...
+            'request_payload' => $data->pendingRequest()->body()?->all(),
+        ];
+
+        if ($data->wasSent()) {
+            $formattedData += [
+                'response_status' => $data->response()->status(),
+                'response_headers' => $data->response()->headers(),
+
+                // TODO: This should be converted to body/content, and stuff like that.
+                'response_body' => $data->response()->json(),
```
Sammyjo20: I think we should just use `->body()` for the raw string body, maybe we could check based on Content-Type too for some light formatting, JSON formatting only perhaps?

src/Debugging/Drivers/ErrorLogDebugger.php
```diff
@@ -0,0 +1,55 @@
...
+    protected function formatData(DebugData $data): array
+    {
+        // TODO: Format the $data, and send it appropriately to the error_log().
+
+        $formattedData = [
+            'method' => $data->method(),
+            'uri' => $data->url(),
+            'request_headers' => $data->pendingRequest()->headers(),
+            'request_query' => $data->pendingRequest()->query()->all(),
+            'request_payload' => $data->pendingRequest()->body()?->all(),
```
Sammyjo20: We've gotta be careful as some bodies like MultipartBody will return an array of MultipartValue objects which may or may not play nicely with debuggers, this is one of the reasons why I wanted to output the sender's response body because that would be the actual string version of what it should be which IMO is better for debugging.

src/Debugging/Drivers/RayDebugger.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        ray($this->formatData($data));
+
+        return $this;
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    protected function formatData(DebugData $data): array
```
Sammyjo20: Maybe move this method into the DebugData DTO so we don't need it on every driver, then if people need to format it further like for Telescope, they do it in the `send` method

src/Traits/HasDebugging.php
```diff
@@ -0,0 +1,38 @@
...
+            $this instanceof Request => $this->middleware(),
+        };
+
+        $callback($debugger = new Debugger);
+
+        $middlewarePipeline->onRequest(function (PendingRequest $pendingRequest) use ($debugger): void {
+            $debugger->send(new DebugData($pendingRequest, null));
+        })->onResponse(function (Response $response) use ($debugger): void {
+            $debugger->send(new DebugData($response->getPendingRequest(), $response));
+        });
```
Sammyjo20: We'll move this into the `registerDefaultMiddleware` method on the PendingRequest so we don't have to worry about the debugger outputting the wrong thing.

Sammyjo20 (commented):
Just left some opening comments and todos for myself, please feel free to reply to any of them if you have any ideas.

I'm still thinking it would be good to get the raw request/response from the sender too, but we'll have to see how feasible that is. I will continue working on this tomorrow night :)

src/Debugging/DebugData.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return $this->pendingRequest->getMethod();
+    }
+
+    public function response(): ?Response
+    {
+        return $this->response;
+    }
+
+    public function statusCode(): ?int
```
juse-less:
```suggestion
    public function getStatusCode(): ?int
```

src/Debugging/DebugData.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return $this->pendingRequest;
+    }
+
+    public function request(): Request
+    {
+        return $this->pendingRequest->getRequest();
+    }
+
+    public function url(): string
```
juse-less:
```suggestion
    public function getUrl(): string
```

src/Debugging/DebugData.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return $this->pendingRequest->getSender();
+    }
+
+    public function pendingRequest(): PendingRequest
+    {
+        return $this->pendingRequest;
+    }
+
+    public function request(): Request
```
juse-less:
```suggestion
    public function getRequest(): Request
```

src/Debugging/DebugData.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return $this->pendingRequest->getConnector();
+    }
+
+    public function sender(): Sender
+    {
+        return $this->pendingRequest->getSender();
+    }
+
+    public function pendingRequest(): PendingRequest
```
juse-less:
```suggestion
    public function getPendingRequest(): PendingRequest
```

src/Debugging/DebugData.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return ! $this->wasSent();
+    }
+
+    public function connector(): Connector
+    {
+        return $this->pendingRequest->getConnector();
+    }
+
+    public function sender(): Sender
```
juse-less:
```suggestion
    public function getSender(): Sender
```

src/Debugging/DebugData.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return ! is_null($this->response);
+    }
+
+    public function wasNotSent(): bool
+    {
+        return ! $this->wasSent();
+    }
+
+    public function connector(): Connector
```
juse-less:
```suggestion
    public function getConnector(): Connector
```

src/Debugging/DebugData.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return $this->pendingRequest->getRequest();
+    }
+
+    public function url(): string
+    {
+        return $this->pendingRequest->getUrl();
+    }
+
+    public function method(): Method
```
juse-less:
```suggestion
    public function getMethod(): Method
```

src/Debugging/DebugData.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return $this->pendingRequest->getMethod();
+    }
+
+    public function response(): ?Response
```
juse-less:
```suggestion
    public function getResponse(): ?Response
```

src/Debugging/DebugData.php
juse-less: Ask, and you shall receive. 😜 
All getters has been prefixed with get in suggestions, so you can pick and choose.

juse-less:
Regarding Telescope, I'm a bit split on this one.
It makes sense to add that driver in the Laravel-specific repo.
But then we'd have scattered drivers.

But it also doesn't make sense to install Telescope as a dev-dependency in this repo - which we should, to be able to test that driver..

Overall, I think it'd be a good idea to add suggestions in composer.json for the various drivers.
Like Ray, and obviously Telescope wherever it ends up.

Sammyjo20:
I agree @juse-less and thank you for working on some of my suggestions 😆 

I'm also thinking that we should release this after v2, I think it's time we finally cut the red tape and let people use V2 as the primary version of Saloon, and then we can release this very quickly after. I don't think we need to introduce any breakin changes.

 I would still like to enrich the debug data with the sender request/response where possible. The main and maybe only reason why is the request body. We could have `pending_request_body` and `raw_body` / `sender_body` on the output as well. This would be super handy, and it would be relatively easy, in our code we add a "mixed" $rawRequest/$senderRequest property and then in our `formatData` method, if the raw request is an instance of `PsrRequest` we can add the additional data.

juse-less:
@Sammyjo20 That is fair enough.
This shouldn't cause any breaking changes, because
1. We don't change any interfaces for anything (other than adding interfaces for the feature itself).
1. The debugging method we add on the default classes already has implementations, and won't randomly start throwing errors when people update Saloon (unless we have a method collision, obviously).

So from that perspective, it _should_ be fine.
I moved the DebuggingDriver interface to src/Contracts, and added a base DebuggingDriver which contains the formatting methods.
I was first gonna add it to the DTO, but.. well.. it's a DTO - it only carries the data.

I'll leave the rest to you, as the remaining of February will be fairly busy for me.
You're free to do whatever you'd like.
Including, but not limited to, copying whatever you'd like from the original branch (in this PR) if you feel it'd be better to start over in a new PR, or so.

src/Debugging/Drivers/RayDebugger.php
juse-less: A DTO only carries the data, so I added the formatting methods on a base driver that all other drivers extend, instead.

src/Debugging/Drivers/ErrorLogDebugger.php
juse-less: I've updated this to first check the Content-Type header, and, depending on that return `response()->json()`, and `response()->xml()`, respectively.
Or, as a default, 'force-call' `response()->json()` (most common), and upon Exception, call `response()->xml()`.
If `response()->xml()` returns false (fail), we'll fallback to raw string.

The reason I try to make an array is that tools like Ray, or even print_r(), that I use for syslog/error log, all present data neater.

Although.. we now have an issue - xml() will return a SimpleXMLElement, so that might be an issue - we'll need to test that.

src/Debugging/Debugger.php
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        $driverName = is_string($driver) ? $driver : $driver->name();
+
+        // We'll validate that the driver exists
+
+        $registeredDriverNames = array_keys($this->getRegisteredDrivers());
+
+        if (! in_array($driverName, $registeredDriverNames, true)) {
+            throw new UnknownDriverException(sprintf('Unable to find the "%s" driver. Registered drivers: %s', $driverName, implode(',', $registeredDriverNames)));
```
juse-less: Just a small output-suggestion.

Current: driver1,driver2,driver3
Suggested: driver1, driver2, driver3


```suggestion
            throw new UnknownDriverException(sprintf('Unable to find the "%s" driver. Registered drivers: %s', $driverName, implode(', ', $registeredDriverNames)));
```

src/Debugging/Debugger.php
```diff
@@ -0,0 +1,227 @@
...
+        $this->registerDriver(new ErrorLogDebugger);
+        $this->registerDriver(new SystemLogDebugger);
+    }
+
+    /**
+     * Register a driver globally
+     *
+     * @param \Saloon\Contracts\DebuggingDriver $driver
+     *
+     * @return $this
```
juse-less:
```suggestion
     * @return static
```

src/Debugging/Drivers/DebuggingDriver.php
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        return $response->body();
+    }
+
+    /**
+     * Format the items of the array into arrays
+     *
+     * @param array $formattedData
+     * @return array
```
juse-less:
```suggestion
     * @param array<string, mixed> $formattedData
     * @return array<string, mixed>
```

src/Debugging/Drivers/DebuggingDriver.php
```diff
@@ -0,0 +1,113 @@
...
+     */
+    protected function formatDataAsArray(array $formattedData): array
+    {
+        return array_map(static function (mixed $item) {
```
juse-less, line 105: Just a suggestion, as I saw no concrete reason for a 'full' Closure.

Interesting that PHPStan doesn't complain about the missing return type on the Closure, though. 🤔
I guess that'll come when we go to level 9.

```suggestion
        return array_map(static fn (mixed $item) => match (true) {
            $item instanceof ArrayStore, $item instanceof BodyRepository => $item->all(),
            $item instanceof MultipartBodyRepository => $item->toArray(),
            default => $item,
        }, $formattedData);
```

src/Traits/HasDebugging.php
```diff
@@ -0,0 +1,34 @@
...
+     * Debugger
+     *
+     * @var \Saloon\Debugging\Debugger|null
+     */
+    protected ?Debugger $debugger = null;
+
+    /**
+     * Retrieve the debugger
+     *
+     * @param callable|null $callback
```
juse-less:
```suggestion
     * @param callable(\Saloon\Debugging\Debugger): (void)|null $callback
```

src/Helpers/Helpers.php
```diff
@@ -71,4 +71,18 @@ public static function value(mixed $value, mixed ...$args): mixed
...
         return $value instanceof Closure ? $value(...$args) : $value;
     }
+
+    /**
+     * Process an item and return it back
+     *
+     * @param mixed $value
+     * @param callable $callback
+     * @return mixed
+     */
```
juse-less:
```suggestion
    /**
     * Process an item and return it back
     *
     * @template TValue
     *
     * @param TValue $value
     * @param callable(TValue): (void) $callback
     * @return TValue
     */
```

src/Debugging/Drivers/TelescopeDebugger.php
```diff
@@ -0,0 +1,54 @@
...
+     * @param \Saloon\Debugging\DebugData $data
+     *
+     * @return void
+     */
+    public function send(DebugData $data): void
+    {
+        Telescope::recordClientRequest(IncomingEntry::make($this->formatData($data)));
+    }
+
+    protected function formatData(DebugData $data): array
```
juse-less:
```suggestion

    /**
     * @param DebugData $data
     * @param bool $asArray
     * @return array<string, mixed>
     */
    protected function formatData(DebugData $data, bool $asArray = false): array
```

juse-less (commented):
This should solve all current PHPStan errors, except for the dependency-related ones.

We spoke about moving the Telescope driver into the Saloon for Laravel repo.
That solves the error here, but we should install Telescope as a dev-dependency there.
The Ray one is solved by installing ray as dev-dependency in this package.

Also added a couple suggestions.

src/Debugging/Drivers/DebuggingDriver.php
```diff
@@ -0,0 +1,113 @@
...
+            return match (true) {
+                $item instanceof ArrayStore, $item instanceof BodyRepository => $item->all(),
+                $item instanceof MultipartBodyRepository => $item->toArray(),
+                default => $item,
+            };
+        }, $formattedData);
```
juse-less: Just a suggestion, as I saw no concrete reason for a 'full' Closure.

Interesting that PHPStan doesn't complain about the missing return type on the Closure, though. 🤔
I guess that'll come when we go to level 9.

```suggestion
        return array_map(static fn (mixed $item) => match (true) {
            $item instanceof ArrayStore, $item instanceof BodyRepository => $item->all(),
            $item instanceof MultipartBodyRepository => $item->toArray(),
            default => $item,
        }, $formattedData);
```

juse-less (commented):
Accidentally clicked the top-row, and not bottom row in the previous review.
Which made GitHub suggest to replace the top row with the entire block, and not all rows.
So I re-did this specific suggestion.

src/Debugging/Debugger.php
Sammyjo20: Great spot

juse-less:
The Ray error in PHPStan is because they can return different versions of Ray, not just the raw one.
We're installing the 'raw' package, and therefore don't have LaravelRay, for instance.

So we'll need to add an exception/ignore in our PHPStan config.

That would resolve everything but Telescope for now. 🙂

juse-less:
https://github.com/Sammyjo20/Saloon/blob/598421c5f2f1d23b39d9dadd22c5c0877c17d2f2/phpstan.dist.neon#L19-L20

@Sammyjo20 See if adding this to phpstan.dist.neon solves it.
```yaml
  ignoreErrors:
  - "#^Unsafe usage of new static\\(\\)\\.$#"
  - "#^Call to method label\\(\\) on an unknown class Spatie\\\\(LaravelRay|RayBundle|WordPressRay|YiiRay)\\\\Ray\\.$#"
```

phpstan.dist.neon
```diff
@@ -17,4 +17,5 @@ parameters:
   - src
 
   ignoreErrors:
-  - "#^Unsafe usage of new static\\(\\)\\.$#"
+   - "#^Unsafe usage of new static\\(\\)\\.$#"
+   - "#^Call to method label\\(\\) on an unknown class Spatie\\\\(LaravelRay|RayBundle|WordPressRay|YiiRay)\\\\Ray\\.$#"
```
juse-less: Looks like an additional space for each line snuck in there.
YAML can be sensitive at times. 😛

```suggestion
  - "#^Unsafe usage of new static\\(\\)\\.$#"
  - "#^Call to method label\\(\\) on an unknown class Spatie\\\\(LaravelRay|RayBundle|WordPressRay|YiiRay)\\\\Ray\\.$#"
```

Sammyjo20:
@juse-less Do you know how Guzzle outputs their debugging? I can't seem to find it here: https://github.com/search?q=repo%3Aguzzle%2Fguzzle%20debug&type=code

juse-less:
@Sammyjo20 https://docs.guzzlephp.org/en/stable/request-options.html#debug

> Set to true or set to a PHP stream returned by fopen() to enable debug output with the handler used to send a request. For example, when using cURL to transfer requests, cURL's verbose of CURLOPT_VERBOSE will be emitted. When using the PHP stream wrapper, stream wrapper notifications will be emitted. If set to true, the output is written to PHP's STDOUT. If a PHP stream is provided, output is written to the stream.

The way I interpret is that,
* if you give it 'true', it'll print to STDOUT.
   So it would depend on SAPI, I guess.
   It does sound reasonable that the CLI SAPI would print to terminal, but not sure about web-based SAPIs.
* if you give it a file resource (`fopen()`, for instance), it'll print the output to that file.

Thinking of the latter, we _should_ be able to use such resource to catch the debug output, and print to our driver of choice.
It'd need some testing, though.

Sammyjo20:
Oh that's really interesting! Thank you for showing me part of the docs there - I'll try making a driver that accepts a resource and see if we can write to it. I assume we'll have to write a JSON-encoded line for each debug log?

Sammyjo20:
This is how the debugging output looks in a file if I provide an `fopen` instance:

```
*   Trying 172.67.157.64:443...
* Connected to tests.saloon.dev (172.67.157.64) port 443 (#0)
* ALPN: offers http/1.1
*  CAfile: /usr/local/etc/openssl@1.1/cert.pem
*  CApath: none
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN: server accepted http/1.1
* Server certificate:
*  subject: CN=*.saloon.dev
*  start date: Feb 22 03:14:59 2023 GMT
*  expire date: May 23 03:14:58 2023 GMT
*  subjectAltName: host "tests.saloon.dev" matched cert's "*.saloon.dev"
*  issuer: C=US; O=Google Trust Services LLC; CN=GTS CA 1P5
*  SSL certificate verify ok.
> GET /api/user HTTP/1.1
Host: tests.saloon.dev
User-Agent: GuzzleHttp/7
Accept: application/json

* old SSL session ID is stale, removing
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Date: Sun, 26 Feb 2023 17:55:43 GMT
< Content-Type: application/json
< Transfer-Encoding: chunked
< Connection: keep-alive
< vary: Accept-Encoding
< Cache-Control: no-cache, private
< x-ratelimit-limit: 1000
< x-ratelimit-remaining: 999
< access-control-allow-origin: *
< x-frame-options: SAMEORIGIN
< x-xss-protection: 1; mode=block
< x-content-type-options: nosniff
< CF-Cache-Status: DYNAMIC
< Report-To: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v3?s=RBa8ufVJ7uKGo0iIc2aVskX7lWZ2u3zATzBUFZwjYcwclKLibYD4f5uho476y9HCbAi0LOAeMtMG7A0LjIzn88hRFTmRKjnu9wdkHBmTqNKVURnFQmuD8unF%2FuBwEK3veYVG"}],"group":"cf-nel","max_age":604800}
< NEL: {"success_fraction":0,"report_to":"cf-nel","max_age":604800}
< Server: cloudflare
< CF-RAY: 79fa8da52d200732-LHR
< alt-svc: h3=":443"; ma=86400, h3-29=":443"; ma=86400
< 
* Connection #0 to host tests.saloon.dev left intact
```

juse-less:
I'd either grab all lines, as essentially a list, or just use it all as an entire string, and put them in a key, 'output'.
DebugData can even contain this information.
We can either add an 'additional info' property, and keep it there, or see if we can fix a centric output that we can put into the DebugData (i.e., regardless of Sender).

```php
$formattedData = [
    // ...
    'output' => [
        // each line
    ],
];
```

Or 
```php
$formattedData = [
    // ...
    'output' => '', // Entire string
];
```

I'm fairly certain that Guzzle _shouldn't_ require a resource opened by `fopen()`, because there's no way to know which function actually opened a resource (`popen()`, `fopen()`, etc).
You can guess based on resource context info, but it's impossible to know.
So, with that in mind, we _should_ be able to give it some other means of resource, that we can stream into our own function/method to grab the actual text output.

I don't have it all in front of me, but.. I _think_ it's fairly easy.
PHPs documentation/manual around streams are a bit scarce, though.

Sammyjo20:
@juse-less I've just made a debug driver that supports `fopen` / resources. Perhaps we need to accept all types of streams but it works really well. This is the output I get. What do you think?

```php
Saloon Request:
{
    "method": "GET",
    "uri": "https:\/\/tests.saloon.dev\/api\/user",
    "request_headers": {
        "Accept": "application\/json"
    },
    "request_query": [],
    "request_payload": null,
    "sender_config": [],
    "request_class": "Saloon\\Tests\\Fixtures\\Requests\\UserRequest",
    "connector_class": "Saloon\\Tests\\Fixtures\\Connectors\\TestConnector",
    "sender_class": "Saloon\\Http\\Senders\\GuzzleSender"
} 
Saloon Response:
{
    "method": "GET",
    "uri": "https:\/\/tests.saloon.dev\/api\/user",
    "request_headers": {
        "Accept": "application\/json"
    },
    "request_query": [],
    "request_payload": null,
    "sender_config": [],
    "request_class": "Saloon\\Tests\\Fixtures\\Requests\\UserRequest",
    "connector_class": "Saloon\\Tests\\Fixtures\\Connectors\\TestConnector",
    "sender_class": "Saloon\\Http\\Senders\\GuzzleSender",
    "response_status": 200,
    "response_headers": {
        "Date": "Sun, 26 Feb 2023 18:12:47 GMT",
        "Content-Type": "application\/json",
        "Transfer-Encoding": "chunked",
        "Connection": "keep-alive",
        "vary": "Accept-Encoding",
        "Cache-Control": "no-cache, private",
        "x-ratelimit-limit": "1000",
        "x-ratelimit-remaining": "999",
        "access-control-allow-origin": "*",
        "x-frame-options": "SAMEORIGIN",
        "x-xss-protection": "1; mode=block",
        "x-content-type-options": "nosniff",
        "CF-Cache-Status": "DYNAMIC",
        "Report-To": "{\"endpoints\":[{\"url\":\"https:\\\/\\\/a.nel.cloudflare.com\\\/report\\\/v3?s=rW8b1QXOA0zSutzo6aTna%2FkVqEdM6j%2FIbeF%2FD2qiq3lbwMYL39E5t%2F%2F%2Ftm0OFHzUMXWYzJZ1DiZTtcwNRzo%2Fr5AeJ6h3XH7gfAwIFvrC8CDJAaQeBVeQYA9c61d%2BgLBbsc4j\"}],\"group\":\"cf-nel\",\"max_age\":604800}",
        "NEL": "{\"success_fraction\":0,\"report_to\":\"cf-nel\",\"max_age\":604800}",
        "Server": "cloudflare",
        "CF-RAY": "79faa6a2e9b9068e-LHR",
        "alt-svc": "h3=\":443\"; ma=86400, h3-29=\":443\"; ma=86400"
    },
    "response_body": {
        "name": "Sammyjo20",
        "actual_name": "Sam",
        "twitter": "@carre_sam"
    },
    "response_class": "Saloon\\Http\\Response"
} 
```

Sammyjo20 (commented):
Self review

Sammyjo20:
@juse-less I'm coming off for the night now but I wrote some pretty decent tests for Ray, although it required a slightly hacky solution of swapping out the ray client. Is this bad practice? I've tried to make it agnositc so if people want to swap out the ray client in their apps they can quite easily, and also so it doesn't look like a bunch of "testing" code in production.

Sammyjo20:
I was also wondering if we should use json_encode for the syslog and error log debuggers? Not necessarily with the JSON_PRETTY_PRINT flag but it’s generally easier to parse / read than var_dump. Thoughts?

src/Debugging/Debugger.php
```diff
@@ -0,0 +1,226 @@
...
+        $registeredDriverNames = array_keys($this->getRegisteredDrivers());
+
+        if (! in_array($driverName, $registeredDriverNames, true)) {
+            throw new UnknownDriverException(sprintf('Unable to find the "%s" driver. Registered drivers: %s', $driverName, implode(', ', $registeredDriverNames)));
+        }
+
+        $this->useDrivers[$driverName] = true;
+
+        return $this;
+    }
```
juse-less: In case there are very niched configs that only applies to a specific driver, it might be useful to be able to individually configure them.
Primarily thinking of the syslog and error log drivers here.
If we're going the JSON route that @Sammyjo20 suggested (which is better, imo), then it'd be nice to be able to set JSON config (like pretty print).

```suggestion
     /**
     * Subscribe to a given driver
     *
     * @param \Saloon\Contracts\DebuggingDriver|string $driver A DebuggingDriver or the name one of a registered 
     one.
     * @param callable(\Saloon\Contracts\DebuggingDriver): (void)|null $configureDriver
     *
     * @return $this
     * @throws \Saloon\Exceptions\UnknownDriverException
     */
    public function usingDriver(DebuggingDriver|string $driver, ?callable $configureDriver = null): static
    {
        if ($driver instanceof DebuggingDriver) {
            $this->registerDriver($driver);
        }

        $driverName = is_string($driver) ? $driver : $driver->name();

        // We'll validate that the driver exists

        $registeredDrivers = $this->getRegisteredDrivers();
        $registeredDrivers[$driverName] ?? throw new UnknownDriverException(sprintf(
            'Unable to find the "%s" driver. Registered drivers: %s',
            $driverName,
            implode(', ', array_keys($registeredDrivers)),
        ));
        
        if (! is_null($configureDriver)) {
            $this->configureDriver($driverName, $configureDriver);
        }

        $this->useDrivers[$driverName] = true;

        return $this;
    }
```

src/Debugging/Debugger.php
```diff
@@ -0,0 +1,226 @@
...
+     * @throws \Saloon\Exceptions\UnknownDriverException
+     */
+    public function onlyDriver(DebuggingDriver|string $driver): static
+    {
+        $this->useDrivers = [];
+
+        $this->usingDriver($driver);
+
+        return $this;
+    }
```
juse-less: Also add the driver config callback here, if we add it to the `usingDriver()` method.

```suggestion
    /**
     * Only use a given driver
     *
     * @param \Saloon\Contracts\DebuggingDriver|string $driver A DebuggingDriver or the name one of a registered one.
     * @param callable(\Saloon\Contracts\DebuggingDriver): (void)|null $configureDriver
     *
     * @return $this
     * @throws \Saloon\Exceptions\UnknownDriverException
     */
    public function onlyDriver(DebuggingDriver|string $driver, ?callable $configureDriver = null): static
    {
        $this->useDrivers = [];

        $this->usingDriver($driver, $configureDriver);

        return $this;
    }
```

src/Debugging/Debugger.php
```diff
@@ -0,0 +1,226 @@
...
+     *
+     * @return $this
+     */
+    public function registerDriver(DebuggingDriver $driver): static
+    {
+        $this->registeredDrivers[$driver->name()] = $driver;
+
+        return $this;
+    }
+
```
juse-less, line 85: In case there are very niched configs that only applies to a specific driver, it might be useful to be able to individually configure them.
Primarily thinking of the syslog and error log drivers here.
If we're going the JSON route that @Sammyjo20 suggested (which is better, imo), then it'd be nice to be able to set JSON config (like pretty print).

```suggestion

    /**
     * Configure an already registered driver.
     *
     * @param string $driverName The name of a registered DebuggingDriver.
     * @param callable(\Saloon\Contracts\DebuggingDriver): (void)|null $callback
     *
     * @return $this
     * @throws \Saloon\Exceptions\UnknownDriverException
     */
    public function configureDriver(string $driverName, callable $callback): static
    {
        $registeredDrivers = $this->getRegisteredDrivers();

        $driver = $registeredDrivers[$driverName] ?? throw new UnknownDriverException(sprintf(
            'Unable to find the "%s" driver. Registered drivers: %s',
            $driverName,
            implode(', ', array_keys($registeredDrivers)),
        ));
        
        $callback($driver);
        
        return $this;
    }

```

juse-less (commented):
@Sammyjo20 I agree that `json_encode()` is better.
That one also utilise the JsonSerializable interface of objects, unlike the current `print_r()`.

I also thought of the pretty print, and that stuff, that you talked about.
What do you think of adding configurability to drivers?
That way we can have non-pretty print by default, but people can configure the syslog and error log drivers to print it that way.
Or is it too bloated?

I guess they're more meant for drivers we register by default.

src/Debugging/Debugger.php
Sammyjo20: I love the idea but I think it _may_ be too much for this version of the debugging, as I'd like to ship this pretty soon. I'll switch the sys_log and error_log drivers to use json_encoding though instead of formatting the data as an array, which also removes the need for that, should I remove the `asArray`?

juse-less: That's a good question.
I _think_ they can be removed?

No concrete reason not to remove comes to mind.

Sammyjo20:
@juse-less I was wondering if adding a "Monolog" driver would add value? Since Symfony/Laravel already have their own implementations it would be quite easy to use a Laravel Log driver in the MonologDebugger

Sammyjo20:
@juse-less I'm pretty happy with going live with this codebase, just want you to give it the final thumbs up too

juse-less:
https://github.com/Sammyjo20/Saloon/pull/165#issuecomment-1436009755

> I opted to use `syslog()`, and `error_log()` directly to not pull in additional dependencies.
> Something like Monolog has a lot more sources, though (including syslog and error log).
> Not sure it's worth it from a debugging perspective, though.

To some extent I can see the use of adding a Monolog driver. Especially to replace our syslog and error log drivers.
But.. most of Monologs channels are more.. chat platforms like Slack, if you know what I mean.
It's obviously far from only chats - but most of them are more related to alerting, rather than debugging.
So from that perspective, I'm not sure it makes sense.
That's why I implemented the syslog and error log drivers separately.

That aside, we can always add a Monolog driver after this has been merged, if there's many people who'd like that.

src/Debugging/Debugger.php
```diff
@@ -0,0 +1,189 @@
...
+     *
+     * @var bool
+     */
+    protected bool $showResponse = false;
+
+    public function __construct()
+    {
+        $this->registerDriver(new RayDebugger);
+        $this->registerDriver(new ErrorLogDebugger);
+        $this->registerDriver(new SystemLogDebugger);
```
juse-less, line 55: I think it might be ideal to somehow very if the driver-requirements are installed.
syslog and error log shouldn't be an issue, but I think we should somehow check if Ray is installed, before registering its driver.

Maybe I'm overthinking it, though.
People would still need to actively choose which driver(s) to use, even if they are registered.
Maybe we should add a new method to the interface like.. `hasPreRequisites()`, or something, in which we check prerequisites like dependencies, and return true/false.
Then in `useDriver()`, we can check if it has the prerequisites, and, if not, throw an Exception that tells the user which prerequisites it needs.

Sammyjo20 (approved):

src/Debugging/Debugger.php
Sammyjo20: Ooooh I really like that! `hasDependencies`?

src/Debugging/Debugger.php
```diff
@@ -0,0 +1,189 @@
...
+
+        $driverName = is_string($driver) ? $driver : $driver->name();
+
+        // We'll validate that the driver exists
+
+        $registeredDriverNames = array_keys($this->getRegisteredDrivers());
+
+        if (! in_array($driverName, $registeredDriverNames, true)) {
+            throw new UnknownDriverException(sprintf('Unable to find the "%s" driver. Registered drivers: %s', $driverName, implode(', ', $registeredDriverNames)));
+        }
```
juse-less: I think this is technically a micro-optimisation, given the realistically low amount of items, but _I feel like_ repetitively sending requests, there's a bit of time to gain here.
Doing a hash lookup should be a lot quicker than `in_array()` the key.
We'd also be using language constructs, rather than function calls.

Throwing Exceptions in expressions has been supported since PHP 8.0.

```suggestion
        // We'll validate that the driver exists

        $registeredDrivers = $this->getRegisteredDrivers();
        $registeredDrivers[$driverName] ?? throw new UnknownDriverException(sprintf(
            'Unable to find the "%s" driver. Registered drivers: %s',
            $driverName,
            implode(', ', array_keys($registeredDrivers)),
        ));
```

Sammyjo20:
@juse-less I agree re your comment about Monolog, the drivers didn't feel very "debuggy".

src/Debugging/Drivers/ErrorLogDebugger.php
```diff
@@ -0,0 +1,28 @@
...
+     * @param \Saloon\Debugging\DebugData $data
+     *
+     * @return void
+     * @throws \JsonException
+     */
+    public function send(DebugData $data): void
+    {
+        $encoded = json_encode($this->formatData($data), JSON_THROW_ON_ERROR);
+
+        error_log(print_r($encoded, true), LOG_DEBUG);
```
juse-less: `json_encode()` already turns it into a string, so no need for `print_r()`.

```suggestion
        error_log($encoded, LOG_DEBUG);
```

src/Debugging/Drivers/SystemLogDebugger.php
```diff
@@ -0,0 +1,28 @@
...
+     * @param \Saloon\Debugging\DebugData $data
+     *
+     * @return void
+     * @throws \JsonException
+     */
+    public function send(DebugData $data): void
+    {
+        $encoded = json_encode($this->formatData($data), JSON_THROW_ON_ERROR);
+
+        syslog(LOG_DEBUG, print_r($encoded, true));
```
juse-less: `json_encode()` already turns it into a string, so no need for `print_r()`.

```suggestion
        syslog(LOG_DEBUG, $encoded);
```

src/Debugging/Debugger.php
Sammyjo20: So what would you imagine we do, should we have a check while registering that checks if "canBeRegistered" or something like that (hasPreRequisites) and then if it doesn't it could just not register that driver? Or we could throw an exception. 

Come to think of it, it seems like quite a big of extra logic. I do agree with you though that having ray registered by default seems a bit odd. Oh I also removed the stream debugger just as I didn't feel it had a place, I've never thought to stream to a file.

src/Http/Connector.php
```diff
@@ -18,8 +19,9 @@
 use Saloon\Traits\Responses\HasCustomResponses;
 use Saloon\Contracts\Connector as ConnectorContract;
 use Saloon\Traits\RequestProperties\HasRequestProperties;
+use Saloon\Contracts\HasDebugging as HasDebuggingContract;
 
-abstract class Connector implements ConnectorContract
+abstract class Connector implements ConnectorContract, HasDebuggingContract
```
juse-less, line 24: I'm a bit unsure about this.
We're using a trait that implements the functionality, but it might be considered a breaking change.
I think right now that it's better to document it, and have people add it to their own connectors?
Connectors always has to be created, regardless, so it's less of a task - albeit annoying for something like this.

So maybe remove both trait and interface, in that case?
I'll let you choose, as I'm really unsure about if this is a breaking change or not.

juse-less (commented):
This is what I found right now - just some thoughts.

If it's not merged tomorrow, I can see if I can give it a less hastily look.

src/Debugging/Debugger.php
juse-less: I think it's better at the `useDriver()` stage.
The reason is because it's only ever there that the user takes an active choice which driver(s) to use for debugging.

src/Debugging/Debugger.php
juse-less: Side note:
Normally you'd store the result of
```php
$registeredDrivers[$driverName] ?? 'something';
```

But since we don't need the result, I opted to omit it.

juse-less:
Imagine adding something like Slack as a debugging driver. Like.. outputting a message in a public channel every time someone debugs a request. 😂

There are channels that still makes sense, like some 'log' services (Loggly, Sentry, etc), APMs, or so.
But I still feel like those would be a bit niched - especially in terms of debugging locally.

src/Contracts/HasDebugging.php
```diff
@@ -0,0 +1,18 @@
...
+namespace Saloon\Contracts;
+
+use Saloon\Debugging\Debugger;
+
+interface HasDebugging
+{
+    /**
+     * Retrieve the debugger
+     *
+     * @param callable|null $callback
```
juse-less: Same as the trait.

```suggestion
     * @param callable(\Saloon\Debugging\Debugger): (void)|null $callback
```

juse-less (commented):
@Sammyjo20 I went through it all again, and this is the only thing I really found.

Normally, I'd probably check dependencies during default registration of drivers, as well.
But it doesn't really give anything, as we'll notify the user if they 'use' that driver.
In this case, if we don't register the driver, it could also give some misleading Exception(s) (driver not registered, instead of lacking dependencies).
I thought I'd write that down, in case you have a thought about it as well.

The only thing apart from my suggestion is the added interface on the base Connector.
I'm still not sure what would be right, so just 'Resolve conversation' if you think it's ok.

src/Http/Connector.php
Sammyjo20: I think this is fine for now as I'm using the contract to detect if we need to use it. Hopefully we can fix the middleware ordering in v3 so we can have your original idea of it all being built into a Saloon plugin.

Sammyjo20 (approved):
I'm happy with the files now!

juse-less:
I am as well. I can't approve, as I started the PR.
But you're ok to merge for me. 🙂

Sammyjo20:
Thanks for your help on this one man, great bit of teamwork as ever 🔥

PR #168, opened 2023-02-23
Fix | Authenticate a PendingRequest after being constructed

### LINKED ISSUE #167: Unable to authenticate a pending request
According to the [docs](https://docs.saloon.dev/digging-deeper/retrying-requests#customising-when-a-retry-is-attempted), I should be able to call 
`$pendingRequest->authenticate(...)` from the retry handler, however it does not seem to be updating the header on the pending request. I have created a reproduction test case [here](https://github.com/Riley19280/Saloon/commit/cd3603d8587406e82961f9aa1f00d424d22b6b86).
```php
test('you can authenticate the pending request inside the retry handler', function () {
    $mockClient = new MockClient([
        MockResponse::make(['name' => 'Sam'], 401),
        MockResponse::make(['name' => 'Gareth'], 200),
    ]);

    $connector = new TestConnector;
    $connector->withMockClient($mockClient);

    $response = $connector->sendAndRetry(new UserRequest, 2, 0, function (Exception $exception, PendingRequest $pendingRequest) {

        $pendingRequest->authenticate(new \Saloon\Http\Auth\TokenAuthenticator('newToken'));

        return true;
    });

    expect($response->status())->toBe(200);
    expect($response->json())->toEqual(['name' => 'Gareth']);
    expect($response->getPendingRequest()->headers()->get('Authorization'))->toEqual('Bearer newToken');
});
```

The test results in `Failed asserting that null matches expected 'Bearer newToken'.`

Sammyjo20:
This PR closes #167.

I think in v3 of Saloon, we should change the `sendAndRetry` method to use a `Request` instance and not a `PendingRequest` instance, similar to the paginators. This means when people need to modify the request, they can do. They could even add special middleware or authenticators because it will be just a cloned request.

For now though, I've implemented a solution which means if you run the `authenticate` method on a fully constructed `PendingRequest` it will run the `set` method right away. It's not ideal but it will be fine for now 🤠

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,8 +33,7 @@
         "pestphp/pest": "^1.21",
         "phpstan/phpstan": "^1.9",
         "spatie/ray": "^1.33",
-        "symfony/dom-crawler": "^6.0",
-        "symfony/stopwatch": "^6.2"
+        "symfony/dom-crawler": "^6.0"
     },
     "suggest": {
         "illuminate/collections": "Required for the response collect() method.",
--- a/src/Http/PendingRequest.php
+++ b/src/Http/PendingRequest.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Saloon\Http;
 
-use Saloon\Contracts\Authenticator;
 use Saloon\Enums\Method;
 use Saloon\Helpers\Config;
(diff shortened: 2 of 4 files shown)


Sammyjo20 (commented):
Self review

PR #170, opened 2023-02-24
Saloon V2 fix: expires_in should be optional

WalrusSoup:
# Fix Expires_In

[The OAuth Spec](https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2) lists `expires_in` as an optional parameter, one which salesforce does not use - which I discovered [here](https://github.com/MyOutDeskLLC/SalesforcePhp/tree/feature/saloon-v2).

Issue:
* OAuth connections that do not use this parameter, despite it being recommended, will cause a `Undefined property: stdClass::$expires_in` error

This PR 
* checks `expires_in` for null and allows `createOAuthAuthenticator` to accept `DateTimeImmutable|null`. 
* adds a new test  `it allows expires_in to be optional` 

I have confirmed all tests still pass with this adjustment.

### DIFF
--- a/src/Traits/OAuth2/AuthorizationCodeGrant.php
+++ b/src/Traits/OAuth2/AuthorizationCodeGrant.php
@@ -173,7 +173,7 @@ trait AuthorizationCodeGrant
 
         $accessToken = $responseData->access_token;
         $refreshToken = $responseData->refresh_token ?? $fallbackRefreshToken;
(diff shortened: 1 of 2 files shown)


Sammyjo20 (approved):
Thank you!

Sammyjo20 (approved):

PR #171, opened 2023-02-24
Feature | Added request modifier to OAuth2 process

Sammyjo20:
This PR adds an optional `requestModifier` property to the access token and refresh token methods to make it easier to tap into the request and make some final adjustments that is vendor specific (like supplying scope again)

```php
$auth->getAccessToken($code, requestModifier: function (Request $request) {
    //
});

$auth->getRefreshToken($code, requestModifier: function (Request $request) {
    //
});

$auth->getUser($code, requestModifier: function (Request $request) {
    //
});
```

This PR also adds a global request modifier for the whole OAuth2 config so you can easily modify any request sent. 

```php
protected function defaultOauthConfig(): OAuthConfig
{
    return OAuthConfig::make()
        ->setClientId('my-client-id')
        ->setClientSecret('my-client-secret')
        ->setRedirectUri('https://my-app.saloon.dev/auth/callback')
	->setRequestModifier(function (Request $request) {
              // $request->headers()->add();
        )},
}
```

@juse-less How did I do on the generics? 😁

### DIFF
--- a/src/Helpers/OAuth2/OAuthConfig.php
+++ b/src/Helpers/OAuth2/OAuthConfig.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Saloon\Helpers\OAuth2;
 
 use Saloon\Traits\Makeable;
+use Saloon\Contracts\Request;
 use Saloon\Exceptions\OAuthConfigValidationException;
 
 /**
@@ -55,6 +56,13 @@ class OAuthConfig
      * @var string
      */
     protected string $userEndpoint = 'user';
+
+    /**
+     * Callable that modifies the OAuth requests
+     *
+     * @var callable(\Saloon\Contracts\Request): (void)|null
+     */
+    protected mixed $requestModifier = null;
 
     /**
      * The default scopes that will be applied to every authorization URL.
@@ -222,6 +230,38 @@ class OAuthConfig
         $this->defaultScopes = $defaultScopes;
 
         return $this;
+    }
+
+    /**
+     * Set the request modifier callable which can be used to modify the request being sent
+     *
+     * @param callable(\Saloon\Contracts\Request): (void) $requestModifier
+     * @return $this
+     */
+    public function setRequestModifier(callable $requestModifier): static
+    {
+        $this->requestModifier = $requestModifier;
+
+        return $this;
+    }
+
+    /**
+     * Invoke the OAuth2 config request modifier
+     *
+     * @param \Saloon\Contracts\Request $request
+     * @return \Saloon\Contracts\Request
+     */
+    public function invokeRequestModifier(Request $request): Request
+    {
+        $requestModifier = $this->requestModifier;
+
+        if (is_null($requestModifier)) {
+            return $request;
+        }
+
+        $requestModifier($request);
+
+        return $request;
     }
 
     /**
--- a/src/Helpers/OAuth2/OAuthConfig.php
+++ b/src/Helpers/OAuth2/OAuthConfig.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Saloon\Helpers\OAuth2;
 
+use Closure;
 use Saloon\Traits\Makeable;
 use Saloon\Contracts\Request;
 use Saloon\Exceptions\OAuthConfigValidationException;
@@ -62,7 +63,7 @@ class OAuthConfig
      *
      * @var \Closure(\Saloon\Contracts\Request): (void)|null
      */
-    protected Closure $requestModifier = null;
+    protected ?Closure $requestModifier = null;
 
     /**
      * The default scopes that will be applied to every authorization URL.
@@ -235,9 +236,7 @@ class OAuthConfig
     /**
      * Set the request modifier callable which can be used to modify the request being sent
      *
-     * @template TRequest of \Saloon\Contracts\Request
-     *
-     * @param callable(TRequest): (void) $requestModifier
+     * @param callable(\Saloon\Contracts\Request): (void) $requestModifier
      * @return $this
      */
     public function setRequestModifier(callable $requestModifier): static
--- a/src/Helpers/OAuth2/OAuthConfig.php
+++ b/src/Helpers/OAuth2/OAuthConfig.php
@@ -60,9 +60,9 @@ class OAuthConfig
     /**
      * Callable that modifies the OAuth requests
      *
-     * @var callable(\Saloon\Contracts\Request): (void)|null
+     * @var \Closure(\Saloon\Contracts\Request): (void)|null
      */
-    protected mixed $requestModifier = null;
+    protected Closure $requestModifier = null;
 
     /**
      * The default scopes that will be applied to every authorization URL.
@@ -235,12 +235,14 @@ class OAuthConfig
     /**
      * Set the request modifier callable which can be used to modify the request being sent
      *
-     * @param callable(\Saloon\Contracts\Request): (void) $requestModifier
+     * @template TRequest of \Saloon\Contracts\Request
+     *
+     * @param callable(TRequest): (void) $requestModifier
      * @return $this
      */
     public function setRequestModifier(callable $requestModifier): static
     {
-        $this->requestModifier = $requestModifier;
+        $this->requestModifier = $requestModifier(...);
 
         return $this;
     }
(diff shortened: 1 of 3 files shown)


src/Traits/OAuth2/AuthorizationCodeGrant.php
```diff
@@ -96,26 +96,35 @@ public function getAuthorizationUrl(array $scopes = [], string $state = null, st
     /**
      * Get the access token.
      *
+     * @template TRequest \Saloon\Contracts\Request
```
juse-less:
```suggestion
     * @template TRequest of \Saloon\Contracts\Request
```

src/Traits/OAuth2/AuthorizationCodeGrant.php
```diff
@@ -96,26 +96,35 @@ public function getAuthorizationUrl(array $scopes = [], string $state = null, st
...
+     *
      * @param string $code
      * @param string|null $state
      * @param string|null $expectedState
      * @param bool $returnResponse
+     * @param callable(TRequest): TRequest|null $requestModifier
```
juse-less:
```suggestion
     * @param callable(TRequest): (void)|null $requestModifier
```

src/Traits/OAuth2/AuthorizationCodeGrant.php
```diff
@@ -129,15 +138,18 @@ public function getAccessToken(string $code, string $state = null, string $expec
     /**
      * Refresh the access token.
      *
+     * @template TRequest \Saloon\Contracts\Request
```
juse-less:
```suggestion
     * @template TRequest of \Saloon\Contracts\Request
```

src/Traits/OAuth2/AuthorizationCodeGrant.php
```diff
@@ -129,15 +138,18 @@ public function getAccessToken(string $code, string $state = null, string $expec
...
+     *
      * @param \Saloon\Contracts\OAuthAuthenticator|string $refreshToken
      * @param bool $returnResponse
+     * @param callable(TRequest): TRequest|null $requestModifier
```
juse-less:
```suggestion
     * @param callable(TRequest): (void)|null $requestModifier
```

juse-less (changes requested):
Reviewing on my phone. Hopefully I got it right myself.

I'd recommend always enclosing callable return types in parentheses.
The reason is that the union and second type hint will otherwise be interpreted as part of the callable return type, not an alternative to the callable.

~~You can use an equal sign for the callable return value, to signal it's optional.~~
But since the returned value would never be used, you should use void instead, imo.
It'll show a more clear intent.

Edit: I believe I actually tested 'optional return types' in my callable type hint PR, and concluded that it's not supported/possible.
So I'm pretty sure that you can't make callable return types optional, and you'll need to type hint something like `mixed|void` instead.

juse-less:
Bugger! Reading the actual code, you never use any returned value from the callables.
So the return type should be void for the callables - not TRequest.

Sammyjo20:
Thank you @juse-less :)

Sammyjo20 (commented):
Self review

juse-less:
@Sammyjo20 Can you wait with merging?
I saw a couple things we should change.

I can add the review in ~30-60 minutes.

Sammyjo20:
Hey @craigpotter if you get a chance would you be able to take a look at this PR please and let me know your thoughts? 😁 Thank you in advance

Sammyjo20 (commented):
Self review

juse-less:
@Sammyjo20 Actually, nvm.

What got me is that the DocBlock states `\Saloon\Contracts\Request`, but we're explicitly using a certain request (like the access token request, and get user request).
For The oAuth config, it does make sense to use the interface, though.
But.. on a second note, as part of the oAuth config, I think we should keep the DocBlocks as is, using request interface.
I think we'd also limit ourselves quite a bit if not.
So.. false alarm. 🙂 


Side note:
I also updated my previous review comment around optional callable return type - please read that, as well. 🙂
It doesn't affect the changes made, but I think I may have spread wrong information. 😛

craigpotter:
> Hey @craigpotter if you get a chance would you be able to take a look at this PR please and let me know your thoughts? 😁 Thank you in advance

Yeah I like it. The Global Request modifier was a nice touch as well. Will already solve the stupid MS Scope issues

src/Helpers/OAuth2/OAuthConfig.php
```diff
@@ -56,6 +57,13 @@ class OAuthConfig
      */
     protected string $userEndpoint = 'user';
 
+    /**
+     * Callable that modifies the OAuth requests
+     *
+     * @var callable(\Saloon\Contracts\Request): (void)|null
+     */
+    protected mixed $requestModifier = null;
```
juse-less: I believe it's better to turn this into a Closure, making it more concrete than 'mixed' (anything).
We can still accept callables on methods, but turn them into Closures.
Using the first-class callable syntax on a callable should also bind them to the class, if it's an invokable class.
So there should be no lost information there.

The reason I don't add a template here, is because it doesn't work, AFAIK.
So we'd need to stick with the raw Request interface as argument.
But we'd need to see what PHPStan feels about once/if everything is applied.

```suggestion
    /**
     * Callable that modifies the OAuth requests
     *
     * @var \Closure(\Saloon\Contracts\Request): (void)|null
     */
    protected Closure $requestModifier = null;
```

src/Helpers/OAuth2/OAuthConfig.php
```diff
@@ -224,6 +232,38 @@ public function setDefaultScopes(array $defaultScopes): static
         return $this;
     }
 
+    /**
+     * Set the request modifier callable which can be used to modify the request being sent
+     *
+     * @param callable(\Saloon\Contracts\Request): (void) $requestModifier
+     * @return $this
+     */
```
juse-less, line 241: We use this in the relevant traits, as well.
As long as the template is used in a parameter, in one way or another, it should be fine.
We'll see if PHPStan thinks otherwise. 👍 
Doing this change, did solve complaints it had for me in earlier PRs, though.

```suggestion
    /**
     * Set the request modifier callable which can be used to modify the request being sent
     *
     * @template TRequest of \Saloon\Contracts\Request
     *
     * @param callable(TRequest): (void) $requestModifier
     * @return $this
     */
```

src/Helpers/OAuth2/OAuthConfig.php
```diff
@@ -224,6 +232,38 @@ public function setDefaultScopes(array $defaultScopes): static
...
+
+        return $this;
+    }
+
+    /**
+     * Invoke the OAuth2 config request modifier
+     *
+     * @param \Saloon\Contracts\Request $request
+     * @return \Saloon\Contracts\Request
+     */
```
juse-less: In this case, we even return the very same request, just possibly mutated.

```suggestion
    /**
     * Invoke the OAuth2 config request modifier
     *
     * @template TRequest of \Saloon\Contracts\Request
     *
     * @param TRequest $request
     * @return TRequest
     */
```

src/Helpers/OAuth2/OAuthConfig.php
```diff
@@ -224,6 +232,38 @@ public function setDefaultScopes(array $defaultScopes): static
...
 
+    /**
+     * Set the request modifier callable which can be used to modify the request being sent
+     *
+     * @param callable(\Saloon\Contracts\Request): (void) $requestModifier
+     * @return $this
+     */
+    public function setRequestModifier(callable $requestModifier): static
+    {
+        $this->requestModifier = $requestModifier;
```
juse-less: Turn into a Closure, if we change the property from mixed to Closure.

```suggestion
        $this->requestModifier = $requestModifier(...);
```

juse-less (commented):
Optional suggestions.
The mixed -> Closure is the change I/we did in the Pool, as well.

src/Helpers/OAuth2/OAuthConfig.php
juse-less: The reason I don't add a template here, is because it doesn't work, AFAIK.
So we'd need to stick with the raw Request interface as argument.
But we'd need to see what PHPStan feels about once/if everything is applied.

juse-less:
Side note on all `@template` - they can be grouped with all other tags, if you want.
I know you group `@param`, `@return`, `@throws`, etc, normally as well.
I'm just used to grouping tags by their name, and didn't realise I had split the templates before now.

src/Helpers/OAuth2/OAuthConfig.php
Sammyjo20: Oh awesome, great suggestion thank you!

PR #177, opened 2023-03-03
Feature | OAuth2 Client Credentials Grant

Sammyjo20:
This PR contains the start of a new OAuth2 grant by popular request - the "Client Credentials" grant. 

I'm still working on this behind the scenes but I have my initial test passes. I'm going to see if I can setup a real OAuth2 server to continue with testing this.


Sammyjo20:
I'm not really sure how the `scope` parameter works with the client credentials grant. I assume it's a string separated by spaces or commas? Or is it usually just a single string?

relaypilot:
Thank you for this PR @Sammyjo20, I'll test it later this week.

About the `scope` parameter, it is always a string where the values are separated by a space.

Example:
```
{
   "scope": "contacts.read contacts.write comments.read comments.write",
   "grant_type": "client_credentials"
}
```

Another example with CURL: 
```
curl --location --request POST 'https://localhost:5000/connect/token' \
--header 'Authorization: Basic myclientid:myclientsecret' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'grant_type=client_credentials' \
--data-urlencode 'scope=account.read account.write'
```

If no scope values are provided, a sensible default should be `all`:
```
{
   "scope": "all",
   "grant_type": "client_credentials"
}
```

Sammyjo20:
Great, @relaypilot that's what I built, however I have added a `scopes` array with a scopes separator where the default is a space. This allows it to be comatible with the `defaultScopes` on the config.

Sammyjo20:
Hey @relaypilot I was just wondering if you have had a chance to test this? 😇

relaypilot:
Hi @Sammyjo20, I did a few tests today, and it works perfectly for my use case.
Happy to merge this 👍🏻

Sammyjo20 (commented):
Self review

Sammyjo20:
I'm just going to write a couple of tests to improve code coverage and I think we're good

boryn:
Hi! We are going to start the OAuth 2.0 integration using Saloon, though are just beginning with this great library. 

Would you add some readme pages to the new branch with examples or just write here a comment on how to use it with OAuth 2.0? I mean how should the flow look like on each of the steps? First getting code based on the scopes, later getting access and refresh tokens, later just refreshing tokens and finally how to make "normal" requests using the access tokens? 

We'd be happy to test this branch :)

Sammyjo20:
Hey @boryn - awesome to hear you're thinking of using Saloon and thank you for the kind words. 

This PR is still in progress but if you are using the normal authorization code flow of OAuth2, I have already documented the process extensively on this page: https://docs.saloon.dev/digging-deeper/oauth2-authentication

If you need the client credentials grant then it will be later this week/this weekend before it's released and the docs are written

Sammyjo20 (commented):
Self review

boryn:
Hi @Sammyjo20!

After digging deeper, I wonder why token refreshing is invoked "manually" like:
```php
if ($authenticator->hasExpired() === true) {
    $authConnector = new SpotifyAuthConnector;
    $authenticator = $authConnector->refreshAccessToken($authenticator);
    
    $user->auth = $authenticator;
    $user->save();
}
```

instead of putting the expiration condition with tokens refresh in the custom authenticator as described here https://docs.saloon.dev/the-basics/authentication - wouldn't it be more automatic? One would not have to think of implementing the expiration check, it would have been handled automatically within the "original" request:

```php
public function set(PendingRequest $pendingRequest): void
    {
        // Here would go logic for optional tokens refresh
        // ...

	// Finally, authenticate the previous PendingRequest before it is sent.
        $pendingRequest->headers()->add('Authorization', 'Bearer ' . $response->json('token'));
    }
```

PS. Please insert a link to "OAuth2 Authentication" on a regular "Authentication" page. I was even not aware that there is a separate "OAuth2 Authentication".

EDIT: Ok, I can see that refreshing of tokens may happen automatically in the model method with authenticated instance which as well is quite convenient. Still not sure if the above solution would not be better and more generic?

PR #181, opened 2023-03-10
Update response with generic template for typehinting DTOS

aaronflorey:
This will allow people to typehint a function with Response<MyDTO>, meaning when they call ->dto() on a response, the IDE will know that it's MyDTO and not mixed.

### DIFF
--- a/src/Contracts/Response.php
+++ b/src/Contracts/Response.php
@@ -10,6 +10,9 @@ use Illuminate\Support\Collection;
 use Psr\Http\Message\StreamInterface;
 use Psr\Http\Message\ResponseInterface;
 
+/**
+ * @template TDto
+ */
 interface Response extends HasHeaders
 {
     /**
@@ -114,7 +117,7 @@ interface Response extends HasHeaders
     /**
      * Convert the response into a DTO
      *
-     * @return mixed
+     * @return TDto
      */
     public function dto(): mixed;
 


Sammyjo20:
Hey @aaronflorey thank you for PRing this - this is something I've really needed for a while! I'm going to tag @juse-less here to take a look too as he's my resident generics expert 🤠

Don't worry about the code style failing, that's normal

juse-less:
The reason we 'can't' do this right now, is seen in the failed PHPStan job.

I've tried tampering back-and-forth to figure out how to add generics to requests and responses to correctly handle it.
While it's nice to be able to define everything on the request, and use a default response, it turned out to be a bit tricky (at least for me) to add generics to the request and correctly propagate to the response.
Especially since we need to add the generic types on each class all over the code base (which is what the PHPStan job complains about).

So as-is, we can't/shouldn't merge this PR.
We need to figure out a proper way to add the generic on all relevant classes, and their types in the entire code base.

Sammyjo20:
Hey @aaronflorey as you can see there are quite a number of PHPStan errors with this change. Do you have any ideas on how we could get around this? 

I agree with Juse here, we should work out how to add it to all classes.

Sammyjo20:
I'm undecided on this one, so thank you for your help @aaronflorey but I am going to close this PR for now.

PR #182, opened 2023-03-13
Feature | Installed parallel testing pest plugin

sneycampos:
Since github workflow runs over a 2-core machine, atleast in "free" tier, we can run the tests using parallel testing


Sammyjo20 (approved):
Thank you!

PR #190, opened 2023-03-25
Feature | Added setUrl and setMethod to PendingRequest

Sammyjo20:
This PR also upgrades to Pest 2.0 🚀


Sammyjo20 (commented):
Self review

PR #195, opened 2023-04-07
Fixed incorrect .editorconfig path/extension expansion

juse-less:
Hey, @Sammyjo20.

I realised I incorrectly added spaces into the path/extension expansion, when adding the NEON/PHPStan stuff for v2.
I've removed the spaces.
So editor config should now correctly set indentation on YAML and NEON files.


Sammyjo20 (approved):
Thank you!

PR #200, opened 2023-04-23
Upgrade | HTTP Message Library to v2.0

Sammyjo20:
- Upgraded psr/http-message to ^2.0
- Deleted bak file


Sammyjo20 (commented):
Self review

PR #203, opened 2023-04-26
Tiny alteration of the fix in #202

juse-less:
Instead of entirely re-binding the wrapped MiddlewarePipeline Closures, we can just use static Closures, which will prevent it from being bound to the specific instance.
We don't use the MiddlewarePipeline inside the wrapper Closures, and this feels less ugly and hacky than my previous fix (#202).

I also added more clarifying comments to the two wrapper-Closures.

I also noticed that the MiddlewarePipeline and the Pipelines are mutable, so we don't actually need to re-assign the properties.
I therefore removed that unnecessary re-assignment.
I hope that was ok.


Sammyjo20 (approved):
Thank you :D

PR #205, opened 2023-04-28
Fix | Rewinds the body if it is seekable

Sammyjo20:
This PR fixes an issue where if you called a body method on a response and then used a different one after, it would be an empty string. This was because the `body()` method wasn't rewinding the underlying stream.

This PR fixes this issue.


Sammyjo20 (commented):
Self review

PR #207, opened 2023-04-30
Fix | Fixed bug with responses not returning the new instance

### LINKED ISSUE #204: Modify response before sending it to the client
Hello,

Is there a way to modify the response before sending it to the client ? I am able to retrieve the response in the middleware using "onResponse()" however I did not find a way to update the content of the body response

Sorry if the question does not belong here

Have a good day!

Sammyjo20:
Fixes #204


Sammyjo20 (commented):
Self review

PR #209, opened 2023-05-08
V3 Feature | Remove Dispatcher

Sammyjo20:
This PR removes the `Dispatcher` class from Saloon as well as removes the `send` and `sendAsync` methods from the `PendingRequest` class as this shouldn't have the ability to send requests in my opinion.


Sammyjo20 (commented):
Self review

Sammyjo20 (commented):
Self review

PR #212, opened 2023-05-09
v3 Feature | Deprecated the HasBody interface

Sammyjo20:


Sammyjo20 (commented):
Self review

Sammyjo20:
Hey @juse-less what are your thoughts on this one? Do you think this is a step back or a step forward? I felt like adding both the interface and the trait was confusing a few developers, and I realise that if we don't include it - we throw an exception anyway, so perhaps we just remove the requirement, check if the `body` method exists, and if it does - we just make sure that it returns the right method.

I'm not sure about this PR or not but I wanted to make it as a consideration for v3

juse-less:
Hey, @Sammyjo20!
Sorry, I forgot to answer this.

I'm somewhat split on removing it.
I don't like it, but I know everyone else does it differently for, simplicity.

I'll try explain my view below, so bear with me.

By removing the interface, the method won't show up in autocompletion, when using interface-based type hinting.

I know everyone always puts the body on the requests, which could be one way to simplify it.
But.. take a HEAD request, for instance - having a body on such a request is a malformed/invalid request.
Which is why I prefer to separate things that doesn't necessarily belong in a certain context.

In my mind, there's 3 'perspectives' in code.
1. Definitions (interfaces)
2. Implementations (classes)
3. Not sure how to classify this one, but.. 're-usable logic' (traits).
    These are used by (2), where applicable, in order to compose the implementation.
    They don't necessarily need to implement a definition, but is a great way to re-use logic around certain topics.

In terms of only relying on the trait, there's a couple things to be aware of.
The first one is the overhead of using functions like `method_exists()`, or `class_uses()` (be it 'user-land' or PHP library) over language constructs like `instanceof`.
The second one is that just because the `method_exists()` finds a method with the name, it doesn't necessarily mean it has the signature you'd expect.
So now you'd end up at least double-checking returned type from that method, unless it's ok for PHP itself to fail somewhere in the logic.

So the way I see it, for both performance and 'type correctness', it's better to keep the interface and run the `instanceof` on the request.

But like I said, I'm probably the only one who doesn't like having a body always available on the request.
So for simplicity for everyone, it's maybe better to ditch both the trait and interface, and just merge them with the request class and interface, respectively?

Let me know if I should clarify anything. 🙂

Sammyjo20:
Thanks @juse-less 

I agree - after a bit more thought, I don't think the interface adds that much cognitive overhead, I mean we also throw an exception if you forget to add the interface. 

With v1 I used to have the body method on every request/connector no matter what, but I found that it confused the developer because they could provide body and not add a trait to instruct Guzzle what the type of data it is, so with v2, if you need body - you should add the interface and trait which then provides the correctly type-hinted `body()` method. It is so much better for developers. 

I think I'll leave it for now and then we can always revisit it if we change our mind.

PR #214, opened 2023-05-13
Cast total results to int before returning

craigpotter:
This PR forces the results of the totalResults method on the paginator to return an int. 

The method already has return type of int however some API don't follow conventions and actually supply this value as a string. 

```json
{
	"ResultInfo": {
		"page": "1",
		"per_page": "20",
		"total_count": "47"
	},
}
```

Without this PR, this results in an error due to a mismatch in the return type.


Sammyjo20 (approved):
Thanks @craigpotter 🦸‍♂️

PR #216, opened 2023-05-14
Fix | Throw FatalRequestException when Guzzle doesn't return a response in the RequestException

### LINKED ISSUE #188: GuzzleSender::createResponse(): Argument #2 ($guzzleResponse) must be of type Psr\Http\Message\ResponseInterface, null given
Hi, I am trying to use the SoloRequest as shown in the [documentation](https://docs.saloon.dev/digging-deeper/solo-requests) but here is what i get : 

image

**Code definition** :
image

**Then** :
image

Am i missing something?

PHP 8.1
Laravel 10.4
Saloon 2.4

Sammyjo20:
Fixes #188 

There's a scenario often found when dealing with Invalid SSL certificates that Guzzle won't throw a `ConnectException`, but it will throw a `RequestException` without a response. Saloon will now check if that response is not provided and if it's null, we will throw a `FatalRequestException`.

### DIFF
--- a/phpstan.baseline.neon
+++ b/phpstan.baseline.neon
@@ -76,7 +76,7 @@ parameters:
 			path: src/Http/Response.php
 
 		-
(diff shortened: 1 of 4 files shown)


Sammyjo20 (commented):
Self review

Sammyjo20 (commented):
Self review

PR #224, opened 2023-05-25
Fix | Added support for Guzzle promises v2

Sammyjo20:
This PR adds support for both v1 and v2 of Guzzle's promise library. This also fixes an issue that some users were having when installing Saloon without a version constraint which would cause people to install Saloon v1 and not v2.


Sammyjo20 (commented):
Self review

PR #229, opened 2023-06-06
Feature | JSON flags on JsonBodyRepository

### LINKED ISSUE #194: Escaped backslashes or unicode in json
Could you provide means to give json parameters when building the body. I would like to not escape my urls within the body

View: https://stackoverflow.com/questions/3723243/why-does-json-encoder-adds-escaping-character-when-encoding-urls/4832057#4832057

Sammyjo20:
Fixes #194 

This PR adds support to customise the JSON flags that are used to encode the JSON body repository in Saloon. For example, if you want to make sure that body that contains slashes doesn't get escaped, you can use the `setJsonFlags` method on the `body` method to overwrite them.

They use a "bitmask" which is slightly different from arrays. For example, to set both `JSON_THROW_ON_ERROR` and `JSON_UNESCAPED_SLASHES` you would need to separate them with a pipe (`|`)

```php
$request = new GetServersRequest;
$request->body()->setJsonFlags(JSON_THROW_ON_ERROR | JSON_UNESCAPED_SLASHES);

$connector->send($request);
```

As Saloon recommends keeping logic standardised inside of requests, I would recommend that you set this within the `__construct` or the `boot` method of the request...

```php
__construct()
{
     $this->body()->setJsonFlags(JSON_THROW_ON_ERROR | JSON_UNESCAPED_SLASHES);
}
```

Side note, @juse-less - any idea why it seems like it's ignoring our baseline file now and spitting out loads of the "ignored" errors?

### DIFF
--- a/phpstan.baseline.neon
+++ b/phpstan.baseline.neon
@@ -16,11 +16,6 @@ parameters:
 			path: src/Http/Connector.php
 
 		-
-			message: "#^Parameter \\#2 \\$ of callable callable\\(\\$this\\(Saloon\\\\Http\\\\Connector\\), TValue\\)\\: void expects TValue, array\\|float\\|int\\|string\\|false\\|null given\\.$#"
-			count: 2
-			path: src/Http/Connector.php
-
-		-
 			message: "#^Parameter \\#1 \\$contents of static method Saloon\\\\Data\\\\RecordedResponse\\:\\:fromFile\\(\\) expects string, bool\\|string given\\.$#"
 			count: 1
 			path: src/Http/Faking/Fixture.php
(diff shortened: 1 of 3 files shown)


Sammyjo20 (commented):
Self review

juse-less:
@Sammyjo20 I believe it's because the lock file isn't checked into the repo, so you'll always get the newest PHPStan version.

They recently fixed a couple things on how they interpret callables, so I reckon that's related.

Sammyjo20:
@juse-less Thanks 👍 I managed to fix it by rebuilding the ignores :)

PR #234, opened 2023-06-09
Make it possible to obfuscate e-mails in recorded fixtures

carnevalle:
Hi

I have just started using Saloon for a project. I noticed that when recording data from API's it would save personal data such as e-mail addresses in the fixtures. This is not always ideal in terms of data security or law (especially in Europa). 

If I fetch a data set containing e-mail addresses which I then commit to Github then I have personal identifiable information in my Github Repo. 

A solution could be to ignore the fixtures and not commit them to Github, but I have tried to make a proof of concept where I obfuscate e-mails.

I have added the following to MockConfig
```php
MockConfig::obfuscateEmails()
```

When this is called it will obfuscate e-mails in fixtures like this:

Before
```json
{
    "users" : [
        {
          "name": "Cristiano Ronaldo",
          "e-mail": "cristiano@example.com"
        },
        {
          "name": "Lionel Messi",
          "e-mail": "lionel@example.com"
        }
    ]
}
```

After
```json
{
    "users" : [
        {
          "name": "Cristiano Ronaldo",
          "e-mail": "c*@e***.com"
        },
        {
          "name": "Lionel Messi",
          "e-mail": "l*@e***.com"
        }
    ]
}
```

I am not sure this is ready for production, but what do you think about something like this?

### DIFF
--- a/src/Data/RecordedResponse.php
+++ b/src/Data/RecordedResponse.php
@@ -7,6 +7,8 @@ namespace Saloon\Data;
 use JsonSerializable;
 use Saloon\Contracts\Response;
 use Saloon\Http\Faking\MockResponse;
+use Saloon\Helpers\Str;
+use Saloon\Helpers\MockConfig;
 
 class RecordedResponse implements JsonSerializable
 {
@@ -67,7 +69,7 @@ class RecordedResponse implements JsonSerializable
         return new static(
             statusCode: $response->status(),
             headers: $response->headers()->all(),
-            data: $response->body(),
+            data: MockConfig::shouldObfuscateEmails() ? Str::obfuscateEmail($response->body()) : $response->body(),
         );
     }
 
--- a/src/Helpers/MockConfig.php
+++ b/src/Helpers/MockConfig.php
@@ -21,6 +21,13 @@ final class MockConfig
     private static bool $throwOnMissingFixtures = false;
 
     /**
+     * Denotes if e-mails in responses should be obfuscated when recording fixtures.
+     *
+     * @var bool
+     */
+    private static bool $obfuscateEmails = false;
+
+    /**
      * Set the fixture path
      *
      * @param string $path
@@ -59,5 +66,25 @@ final class MockConfig
     public static function isThrowingOnMissingFixtures(): bool
     {
         return self::$throwOnMissingFixtures;
+    }
+
+    /**
+     * Obfuscate e-mails in responses when recording fixtures
+     *
+     * @return void
+     */
+    public static function obfuscateEmails(): void
+    {
+        self::$obfuscateEmails = true;
+    }
+
+    /**
+     * Should we obfuscate e-mails in fixtures?
+     *
+     * @return bool
+     */
+    public static function shouldObfuscateEmails(): bool
+    {
+        return self::$obfuscateEmails;
     }
 }
--- a/src/Helpers/Str.php
+++ b/src/Helpers/Str.php
@@ -102,4 +102,28 @@ class Str
 
         return $string;
     }
+    
+    // Sources
+    // https://stackoverflow.com/questions/68639402/looking-for-regex-to-obfuscate-email-to-tw-de
+    // https://github.com/gremo/email-obfuscator
+    public static function obfuscateEmails($string)
+    {
+        $patterns = array(
+            '|[_a-z0-9-]+(?:\.[_a-z0-9-]+)*@[a-z0-9-]+(?:\.[a-z0-9-]+)*(?:\.[a-z]{2,3})|i', // plain emails
+        );
+
+        foreach ($patterns as $pattern) {
+            $string = preg_replace_callback($pattern, function ($parts) {
+                // Clean up element parts.
+                $parts = array_map('trim', $parts);
+
+                $obfuscate_pattern = "/^\w\K[^\s@]+@(\w)[^\s.@]+/";
+                $replacement = "*@$1***";
+                return preg_replace($obfuscate_pattern, $replacement, $parts[0]);
+
+            }, $string);
+        }
+
+        return $string;
+    }    
 }


Sammyjo20:
Hey @carnevalle 

Many thanks for this PR - this is an awesome idea 🔥

I do however want to expand a little on it. Emails won't be the only thing inside of fixtures that could potentially be sensitive, and while this would work to obfuscate emails - what about names, addresses etc?

I wonder how we could do it for other properties. Perhaps an array of keys should be randomized? Not quite sure the best way to go... What are your thoughts?

```php
MockResponse::fixture('some-fixture', hideAttributes: [
      'data' => [
            '*' => [
                  'first_name',
                  'last_name',
                  'email',
             ],
      ],
]);
```

Not really a fan of my idea above because it requires having to define something the first time its recorded.

carnevalle:
Great you like it! :-)

I actually had i similar thought after I submitted the PR as a lot of other information could be sensitive, but unfortunately not as easy to identify as e-mail adresses :-)

On top of my head I can think of 
- IP addresses
- ID's
- Names
- Phone numbers
- Addresses
- Tokens
- etc.

So it needs to be controlled by the user which properties it should handle. 

I also thought if it would be an idea to user Faker instead of obfuscation. In this way it remove sensitive data and create readable test-data instead of just hiding stuff.  

I am not super sure about the best way to achieve this, but perhaps it could be possible to define which data should be translated for testing inside a Request? 

In this example you would define the key and what Faker function should be used to replace the data.

```php
public function replacementRules()
{
    return [
        'email' => 'email',
        'first_name' => 'firstName',
        'ip' => 'ipv4'
    ];
}
```

I could try to see if I could get something to work after the weekend.

Sammyjo20:
@carnevalle I will look forward to seeing what you propose and we can work together to find a good solution.

ariaieboy:
Can't we introduce something like laravel factories?

Users should define a factory class for their Requests. 

They can return a simple array as a result, like the `Mock Responses` 

Or, by using a trait, they can tell the factory to use `Recording Responses` to fetch the result. Then users can modify the result in any way they want using the factory class and return the cleaned result from the factory.

Sammyjo20:
@ariaieboy thats actually a really great idea. You could create a MockResponse factory which could have various methods either like “defineFixtureName” or “defineBody” (for manual) and then you could have “defineSensitiveParameters” method which could return an array?

ariaieboy:
> @ariaieboy thats actually a really great idea. You could create a MockResponse factory which could have various methods either like “defineFixtureName” or “defineBody” (for manual) and then you could have “defineSensitiveParameters” method which could return an array?

Exactly.
about the sensitive parameters, we can create a method like `mutateBody($body)` that gets the defineBody or the retrieved body from the API and users have full control of the body before it goes to the test. 
And we can add a function that uses Faker rules to replace the sensitive data automatically like what @carnevalle suggested above:
```php
public function replacementRules()
{
    return [
        'email' => 'email',
        'first_name' => 'firstName',
        'ip' => 'ipv4'
    ];
}
```
But instead of rules I think we should use the exact value that users want to be used for the given key.
```php
public function replacementRules()
{
    return [
        'email' => $faker->email(),
        'first_name' => 'my fake name',
        'ip' => '192.168.1.1'
    ];
}
```

Sammyjo20:
Hey @carnevalle @ariaieboy I've come up with something that I think would be fairly simple to implement, and happy to make a PR for v2. I think we could introduce "Fixture Classes" which allow you to define the fixture path but also some methods to allow for modification of the data. I reckon something like this...

```php
class ClientsFixture extends Fixture
{
     public function defineFixture(): string
     {
         return 'clients'; // Normal fixture name
     }

     public function swapSensitiveJsonParams(): array
     {
           return [
                   'email' => '***',
                   'last_name' => $this->faker->last_name // Or use faker
           ];
     }
}
```

Then it can be used like this:

```php
new MockClient([
      SomeRequest => ClientsFixture::class,
      // or? MockResponse::fixture(ClientsFixture::class)
])
```

The idea is that whenever the code spots an "email" object it will replace it with what you define. Do you think we need anything else? I was wondering if there was anything else we need like wildcards or anything?

carnevalle:
I think this solution would be sufficient - at least for my use case. I think it would be cool if it would generate a different e-mail if using Faker for each occurence in data set, so the `swapSensitiveJsonParams` would have to run on each find.

I was playing around with this function if it is useful in any way. I am not a my dev computer right now so can't check that the code still runs :-)

```php
function json_replace($key, $json)
{
    $faker = Faker\Factory::create();
    $pattern = '/"' . $key . '":".*?"/';
    $json= preg_replace_callback($pattern, function ($parts) use ($pattern, $key, $value, $faker){

        if($key == 'name'){
            $value = $faker->name;
        }else if($key == 'e-mail'){
            $value = $faker->email;
        }else{
            $value = $faker->ipv4;
        }

        $replacement = '"' . $key . '":"' . $value . '"';
        return preg_replace($pattern, $replacement, $parts[0]);

    }, $json);

    return $json;
}

$data = [
    'users' => [
        [
            'name' => 'Lionel Messi',
            'e-mail' => 'lionel@example.com',
            'ip' => '192.0.0.1'
        ],
        [
            'name' => 'Cristiano Ronaldo',
            'e-mail' => 'cristiano@example.com',
            'ip' => '192.0.0.1'
        ]
    ],
    'result' => [
        'users' => [
            [
                'name' => 'Lionel Messi',
                'e-mail' => 'lionel@example.com'
            ],
            [
                'name' => 'Cristiano Ronaldo',
                'e-mail' => 'cristiano@example.com'
            ]
        ]
    ]
];

$json = json_encode($data);

$json = json_replace('name', $json);
$json = json_replace('e-mail', $json);
$json = json_replace('ip', $json);

return $json;
```

Sammyjo20:
Hey @carnevalle thanks for the code suggestion. I've actually had quite a lot of fun with this tonight and I've got this working so far - both header and JSON body redaction. You can replace it with anything, including faker if you wish. It's recursive too so it will catch nested objects.

```php
class SafeUserFixture extends Fixture
{
    /**
     * Define the name of the fixture
     *
     * @return string
     */
    protected function defineName(): string
    {
        return 'user';
    }

    /**
     * Define the sensitive headers
     *
     * @return array
     */
    protected function defineSensitiveHeaders(): array
    {
        return [
            'Server' => 'secret',
            'Cache-Control' => function ($value) {
                return $value . ', yeehaw';
            },
        ];
    }

    /**
     * Swap any sensitive JSON parameters
     *
     * @return array
     */
    protected function defineSensitiveJsonParameters(): array
    {
        return [
            // You can also define callables that should be run to replace the value!

            'name' => static function (string $value) {
                return substr_replace($value, 'xxx', 1);
            },
            'actual_name' => 'REDACTED',
            'twitter' => '@saloonphp'
        ];
    }
}
```

Sammyjo20:
I'm just working on another method which allows you to write your own redaction logic, so in your case if you wanted to look for any value that looked like an email, you could. Perhaps I can do match values too.

Sammyjo20:
@carnevalle I think with my implementation you could use a closure and then call faker inside of that closure to generate a unique fake email every time :D

Sammyjo20:
Thanks for contributing @carnevalle but I have now merged #244 which should have all the features you need. I'll write up the docs in a sec!

carnevalle:
Great job again @Sammyjo20  :-)

PR #236, opened 2023-06-15
Feature | HasStreamBody

Sammyjo20:
This PR introduces a new body type for Saloon's POST requests - `HasStreamBody`! With this, you will be able to specify a stream to be sent to your APIs. This could be the resource from `fopen` or even a stream that implements the `StreamInterface`.


Sammyjo20 (commented):
Self review

PR #237, opened 2023-06-15
Fix | Unset the sender when serializing

Sammyjo20:
This PR introduces a `__sleep()` function on the connector which will destroy the sender instance when attempting to serialize. This is because any open connections that a sender may have must be removed when serializing. 

There was an issue where if the sender was used, then you wouldn't be able to serialize the sender because Guzzle uses closures inside of its handlers.


Sammyjo20 (commented):
Self review

PR #242, opened 2023-06-23
Feature | Ability to overwrite OAuth2 request classes

### LINKED ISSUE #239: GetAccessTokenRequest with HasJsonBody
Hello!

This is an amazing package that's really a pleasure to work with :)

The problem I'm running into is with the **getAccessToken()** method. Looking into this method, I can see that it creates a **GetAccessTokenRequest** which uses the **HasFormBody** trait. This sets the content type for the token request to "application/x-www-form-urlencoded". In my case, I need the posted content type to be JSON as this format is required by the API so that it will accept the post. 

If I change the **GetAccessTokenRequest** to use the **HasJsonBody**, the request posts correctly. Would it be possible to have the **GetAccessTokenRequest** be capable of using a JSON content type as well? Or should I just build out a new trait based on **AuthorizationCodeGrant**?

I was also thinking that maybe it would be good to have the getAccessToken method accept a request strategy so that I could use a different one that uses HasJsonBody. I dunno, just throwing it out there :)

Thanks!

**Quick update:**
The way I've solved this for the moment, is to create a new trait that extends the **AuthorizationCodeGrant** trait and overrides the getAccessToken so that I can use a custom request that uses **HasJsonBody**.

I'm thinking there might be a better way within the framework, but that's just where I'm at now,

Sammyjo20:
Fixes #239

This PR adds a few overwritable protected methods on the `AuthorizationCodeGrant` and `ClientCredentialsGrant` OAuth2 traits. These methods will allow you to overwrite the instantiation of the request classes yourself, so if you need to completely swap the classes out to your own requests, you can now do it.

With future versions of Saloon, I think this should be the preferred way of modifying the request classes, including adding query parameters or changing the request body, because now we have multiple ways of modifying the requests and it may get a little confusing.

### DIFF
--- a/src/Traits/OAuth2/AuthorizationCodeGrant.php
+++ b/src/Traits/OAuth2/AuthorizationCodeGrant.php
@@ -8,8 +8,10 @@ use DateTimeImmutable;
 use Saloon\Helpers\Str;
 use Saloon\Helpers\Date;
 use InvalidArgumentException;
+use Saloon\Contracts\Request;
 use Saloon\Helpers\URLHelper;
 use Saloon\Contracts\Response;
+use Saloon\Helpers\OAuth2\OAuthConfig;
 use Saloon\Http\OAuth2\GetUserRequest;
 use Saloon\Contracts\OAuthAuthenticator;
 use Saloon\Exceptions\InvalidStateException;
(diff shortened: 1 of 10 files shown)


Sammyjo20 (commented):
Self review 🚀

PR #243, opened 2023-06-23
Feature | V3 Implement More PSR

Sammyjo20:
This PR helps step Saloon step further away from a client implementation (like Guzzle). 🚀

Previously, the PendingRequest would go to the sender class, and then it would be the sender's responsibility to build up the request and send it properly. With this PR - Saloon takes the strain and dependency off the sender by building its own PSR-7 requests and streams for all the body types. This allows Saloon to be more independent of the implementation and potentially remove Guzzle as a dependency in later versions. 

This PR also includes a few other nice changes like renaming `SimulatedResponsePayload` (woof!) to `FakeResponse` which is much more readable in my opinion. 

This PR also introduces a minimum requirement for SSL to TLS 1.2. 

## Features
- Renamed `SimulatedResponsePayload` to `FakeResponse`
- Minimum SSL version of TLS 1.2 (can be overwritten)
- Saloon now handles PSR-7 requests 
- `handlePsrRequest` event hook on connector/request to make final amendments to the PSR-7 request (For special edge-cases)
- Minimum PHP 8.2
- Basic auth is now implemented by ourselves
- Moved delay into Saloon codebase rather than Guzzle

## New PSR Methods
With this PR, there are a few new PSR methods which can now be called on the `PendingRequest` class.
- `createPsrRequest()`
- `getUri`

## Todo
- [ ] Extract GuzzleSender into the GuzzleSender library (After this PR)


Sammyjo20 (commented):
Self review

Sammyjo20 (commented):
Self review

Sammyjo20 (commented):
Self review

PR #244, opened 2023-06-24
Feature | Swap Sensitive Properties Within Fixtures

Sammyjo20:
This PR introduces a really cool new addition to fixtures - the ability to redact and remove sensitive information from them! This PR allows you to replace any of the headers or any JSON body with any of your custom redacted text. You can also provide a callback if you would like to modify the data like leave it half-complete and redact the rest. 

From this PR, you will be able to create "Fixture" classes which extend Saloon's base fixtures.

```php
class SafeUserFixture extends Fixture
{
    /**
     * Define the name of the fixture
     *
     * @return string
     */
    protected function defineName(): string
    {
        return 'user';
    }

    /**
     * Define the sensitive headers
     *
     * @return array
     */
    protected function defineSensitiveHeaders(): array
    {
        return [
            'Server' => 'secret',
            'Cache-Control' => function ($value) {
                return $value . ', yeehaw';
            },
        ];
    }

    /**
     * Swap any sensitive JSON parameters
     *
     * @return array
     */
    protected function defineSensitiveJsonParameters(): array
    {
        return [
            // You can also define callables that should be run to replace the value!

            'name' => static function (string $value) {
                return substr_replace($value, 'xxx', 1);
            },
            'actual_name' => 'REDACTED',
            'twitter' => '@saloonphp'
        ];
    }
}
```

This is off the back of the PR #234 where the idea was to be able to find and replace any emails that match a regex.

I wonder if there should still be able to find a way to match header values and JSON body from a given regex? Would be happy to discuss this.

> **Note**
>  The first response (original) is not redacted, only the subsequent responses are redacted.

## Todo
- [x] Add the ability to redact based on Regex on the values


Sammyjo20 (commented):
Self review

Sammyjo20:
Hey @carnevalle @ariaieboy

Would you mind having a look through this and letting me know your thoughts on the implementation and developer outcome? I'd also like to look into potentially replacing values based on regex. Do you think the regex searching would add value? If so feel free to discuss how we could implement it here.

Sammyjo20:
Also small caveat - the first response (original) is not redacted, only the subsequent responses are redacted.

ariaieboy:
We need some global filtering on values. It can use regex or simple text replacement.
It's useful when we work with non-structured data. For example, in caprover, we can retrieve app logs using API, and the records can contain sensitive data, for example, emails or private URLs, etc.

Sammyjo20:
I absolutely agree. We sometimes deal with CSV files sent over APIs so its hard to work out the structure to replace sensitive data. The regex for things like emails, IP addresses, URLs etc is going to be fine but it'll be a lot harder to work out replacements for first names, last names etc.

ariaieboy:
> I absolutely agree. We sometimes deal with CSV files sent over APIs so its hard to work out the structure to replace sensitive data. The regex for things like emails, IP addresses, URLs etc is going to be fine but it'll be a lot harder to work out replacements for first names, last names etc.

That's the programmer's problem. We only need to add regex support.

Sammyjo20:
@ariaieboy I know the MVP would be useful to have regex support, but I do think having the ability to select specific attributes to obscure is also useful - especially when dealing with JSON, the same with headers. 

Do you think what I implemented adds value? 

I'll look at doing regex support as well, but I don't want to clutter the codebase

ariaieboy:
> @ariaieboy I know the MVP would be useful to have regex support, but I do think having the ability to select specific attributes to obscure is also helpful - especially when dealing with JSON, the same with headers. 
> 
> Do you think what I implemented adds value? 
> 
> I'll look at doing regex support as well, but I don't want to clutter the codebase

 The current implementation is perfect. I need another function that runs a simple preg_replace on all values so we can filter using regex. Globally.

Sammyjo20:
Apologies for the delay on this one guys. I am still interested in working on this PR but taking a little break from the open source work for a couple of weeks, looking forward to jumping back in soon!

Sammyjo20:
Hey @carnevalle @ariaieboy

I hope you both are well :) I have now completed this PR and have added another method `defineSensitiveRegexPatterns` on the Fixture class. Here you can define an array of patterns as well as a value to replace it with, or a callback which will be invoked at the time of replacing. 

```php
protected function defineSensitiveRegexPatterns(): array
{
    return [
        // Twitter Handle
        '/(\@)([a-z0-9_]{1,15})/i' => '**REDACTED-TWITTER**',
        // The name Sam
        '/Sam/' => 'Taylor',
    ];
}
```

Sammyjo20 (commented):
Self review

Issue #245, opened 2023-06-26
Using `MockClient` in your app, and not in tests, seems to append 'public' to the fixtures path.


danmatthews:
I was using the mockclient to get some base fixtures while making the actual calls in my app running in the browser and it appears to have appended 'public' to the start of the fixtures path:

`public/tests/Fixtures/Saloon/{myfixture}`

I haven't had time to dive in to submit a PR, but i'm guessing that it's using the `__DIR__` or script base path, which might be causing this?

This also happens even if you explicitly set the path using `MockConfig`:

```
// Still happens
MockConfig::setFixturePath('tests/Fixtures/Saloon');
```

danmatthews:
`vendor/sammyjo20/saloon/src/Http/Faking/Fixture.php:46`

```
// Called in tinkerwell, produces: "tests/Fixtures/Saloon"
(new Saloon\Helpers\Storage(\Saloon\Helpers\MockConfig::getFixturePath()))->getBaseDirectory();
```

Called in the browser, produces:

<img width="303" alt="CleanShot 2023-06-26 at 11 53 22@2x" src="https://github.com/Sammyjo20/Saloon/assets/149426/cbd1db75-3c79-47f2-b865-ff114681905b">

Which, should be fine!

danmatthews:
But yet weirdly still produces the output at `public/`:

<img width="489" alt="CleanShot 2023-06-26 at 11 54 49@2x" src="https://github.com/Sammyjo20/Saloon/assets/149426/638ed074-286f-407d-9473-7b63430cdaf2">

danmatthews:
I think it's: `vendor/sammyjo20/saloon/src/Helpers/Storage.php:112`

`file_put_contents` is probably operating within `public/` ,might be better to move to use 

`file_put_contents(base_path($fullPath), $contents));` or something similar? Will try and test when I can.

danmatthews:
<img width="388" alt="CleanShot 2023-06-26 at 12 08 46@2x" src="https://github.com/Sammyjo20/Saloon/assets/149426/c8cce24e-530b-4132-99aa-869f6d472dc2">

`base_path` fixes it at least for this instance, will pull a copy of the package, PR + run tests in a second.

Sammyjo20:
[...]

PR #246, opened 2023-06-26
[WIP] Update Storage.php

danmatthews:
Add `base_path` to the storage helper to prevent overspilling into the public directory.

Sorts #245 if needed.

Obviously the downside to this is that you can only use directories within your laravel base install, depends if that's an okay trade off.

### DIFF
--- a/src/Helpers/Storage.php
+++ b/src/Helpers/Storage.php
@@ -31,7 +31,7 @@ class Storage
             $createMissingBaseDirectory ? $this->createDirectory($baseDirectory) : throw new DirectoryNotFoundException($baseDirectory);
         }
 
-        $this->baseDirectory = $baseDirectory;
+        $this->baseDirectory = base_path($baseDirectory);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


danmatthews:
Oh poop, never thought about framework-agnostic use case, let me take a look.

danmatthews:
Obviously, the alternative to this is to you just use `base_path` in the `MockConfig::setFixturePath('tests/other-directory')` call - might be good to point out this to users in the docs?

Sammyjo20:
Hey @danmatthews many thanks for the PR 🙌 I think rather than changing the code to detect for specific frameworks, I think it would be better to update the documentation, or maybe even change the Saloon-Laravel package to update the `setFixturePath` inside of the service provider to use the `base_path`. 

I'd be happy to accept a PR on the saloon-laravel side if you are happy to work on that, otherwise I can put a disclaimer on the docs. 

Cheers!
Sam

danmatthews:
Cheers @Sammyjo20 - sorry, i started all this then as soon as I solved my problem I got tasked with something else.

Maybe a hint in the docs is all it would take!

PR #251, opened 2023-07-03
Feature | Added `SoapClientSender`

aydinfatih:
This PR introduces a new sender type for Saloon's SOAP requests - `SoapClientSender`! With this, you will be able to specify a `SoapClient` request to be sent to your Web Services.

### Added 
- Added `HttpResponse` contract
- Added `SoapResponse`
- Added `ext-soap` dependency to `composer.json`
- Added `SoapClientSender`

### Changed
- Updated phpstan.baseline.neon
- Changed `ResponseContract` to `HttpResponseContract`
- Moved Response to Responses/Response
- Removed `close` method from `Response`

### Test
- Added `wsdlUrl()` function to Pest
- Added `SoapRequest` for test
- Added `SoapClientConnector` for test

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,8 @@
         "php": "^8.1",
         "guzzlehttp/guzzle": "^7.5",
         "guzzlehttp/promises": "^1.5 || ^2.0",
-        "psr/http-message": "^1.1 || ^2.0"
+        "psr/http-message": "^1.1 || ^2.0",
+        "ext-soap": "*"
     },
     "require-dev": {
         "ext-simplexml": "*",
--- a/phpstan.baseline.neon
+++ b/phpstan.baseline.neon
@@ -46,21 +46,6 @@ parameters:
 			path: src/Http/PendingRequest.php
 
 		-
-			message: "#^Parameter \\#1 \\$callback of function call_user_func_array expects callable\\(\\)\\: mixed, \\(callable\\(\\)\\: mixed\\)\\|object given\\.$#"
-			count: 2
-			path: src/Http/Response.php
-
-		-
-			message: "#^Parameter \\#1 \\$object of method ReflectionMethod\\:\\:invoke\\(\\) expects object\\|null, object\\|string given\\.$#"
-			count: 1
-			path: src/Http/Response.php
-
-		-
-			message: "#^Parameter \\#1 \\$objectOrClass of class ReflectionClass constructor expects class\\-string\\<T of object\\>\\|T of object, object\\|string given\\.$#"
-			count: 1
-			path: src/Http/Response.php
-
-		-
 			message: "#^Method Saloon\\\\Http\\\\Senders\\\\GuzzleSender\\:\\:createRequestOptions\\(\\) should return array\\<'allow_redirects'\\|'auth'\\|'body'\\|'cert'\\|'connect_timeout'\\|'cookies'\\|'crypto_method'\\|'debug'\\|'decode_content'\\|'delay'\\|'expect'\\|'force_ip_resolve'\\|'form_params'\\|'headers'\\|'http_errors'\\|'idn_conversion'\\|'json'\\|'multipart'\\|'on_headers'\\|'on_stats'\\|'progress'\\|'proxy'\\|'query'\\|'read_timeout'\\|'sink'\\|'ssl_key'\\|'stream'\\|'synchronous'\\|'timeout'\\|'verify'\\|'version', mixed\\> but returns array\\<string, mixed\\>\\.$#"
 			count: 2
 			path: src/Http/Senders/GuzzleSender.php
--- a/src/Contracts/HttpResponse.php
+++ b/src/Contracts/HttpResponse.php
@@ -1,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Saloon\Contracts;
+
+use Throwable;
+use Psr\Http\Message\StreamInterface;
+use Psr\Http\Message\ResponseInterface;
+
+interface HttpResponse extends Response
+{
+    /**
+     * Create an instance of the response from a PSR response
+     *
+     * @param \Saloon\Contracts\PendingRequest $pendingRequest
+     * @param \Psr\Http\Message\ResponseInterface $psrResponse
+     * @param \Throwable|null $senderException
+     * @return $this
+     */
+    public static function fromPsrResponse(ResponseInterface $psrResponse, PendingRequest $pendingRequest, ?Throwable $senderException = null): static;
+
+    /**
+     * Create a PSR response from the raw response.
+     *
+     * @return \Psr\Http\Message\ResponseInterface
+     */
+    public function getPsrResponse(): ResponseInterface;
+
+    /**
+     * Get the body as a stream. Don't forget to close the stream after using ->close().
+     *
+     * @return \Psr\Http\Message\StreamInterface
+     */
+    public function stream(): StreamInterface;
+
+    /**
+     * Close the stream and any underlying resources.
+     *
+     * @return $this
+     */
+    public function close(): static;
+}
--- a/src/Contracts/Response.php
+++ b/src/Contracts/Response.php
@@ -7,28 +7,9 @@ namespace Saloon\Contracts;
 use Throwable;
 use SimpleXMLElement;
 use Illuminate\Support\Collection;
-use Psr\Http\Message\StreamInterface;
-use Psr\Http\Message\ResponseInterface;
 
 interface Response extends HasHeaders
 {
-    /**
-     * Create an instance of the response from a PSR response
-     *
-     * @param \Saloon\Contracts\PendingRequest $pendingRequest
-     * @param \Psr\Http\Message\ResponseInterface $psrResponse
-     * @param \Throwable|null $senderException
-     * @return $this
-     */
-    public static function fromPsrResponse(ResponseInterface $psrResponse, PendingRequest $pendingRequest, ?Throwable $senderException = null): static;
-
-    /**
-     * Create a PSR response from the raw response.
-     *
-     * @return \Psr\Http\Message\ResponseInterface
-     */
-    public function getPsrResponse(): ResponseInterface;
-
     /**
      * Get the underlying PendingRequest that created the response
      *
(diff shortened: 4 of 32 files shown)


deligoez:
Fantastic work on adding SOAP support! This is precisely the feature I've been needing. Great job! 👏🔥

Sammyjo20:
Hey @aydinfatih 

Many thanks for creating this PR, I really appreciate the effort that you have put in to help get Saloon to work with SOAP requests. I wasn't sure if it was a big desire for many people, but since you have started the work on it I think it would be valuable to consider pushing forward.

There are a few things I have seen when going through the code and a few considerations.

- Could you write a test for sending a request synchronously / asynchronously using this sender? 
- Could you also write some tests to ensure the right endpoints are being built? 

Additionally, I think we need to consider some things

- I think it's cool that you're using the headers in Saloon to create the SoapHeaders, but what about HTTP headers? What if we need to send a custom HTTP header? Can this be done with the SoapClient, and I wonder how we could do it in Saloon? I think a good solution could be that if the value provided inside of the headers array isn't a `SoapHeader`, then we treat it like a HTTP header.
- I also think perhaps we need a new abstract request class called `SoapRequest` or something like that where we alias the `resolveEndpoint` method to `resolveAction` or something similar? As that is the actual action that is being invoked from the WSDL file in the base URL?
- How come you have renamed `Response` to `HttpResponse`? Additionally, the changes to the interface would result in a breaking API change, so please could you leave any breaking changes to the interfaces. I'm happy if we just have one Response class and not separate SoapResponse and HttpResponse. This is because as explained below, we might be able to create a PSR response from the SoapResponse.
- I also wonder if we can perhaps create a PSR response from the SoapClient's response, so we don't have to worry about removing the `close` capabilities and things like that. You can use Guzzle's `Utils` class. It has a useful `Utils::streamFor()` method where you can create a stream interface from a string.


What are your thoughts?

Optionally, if these changes you would like me to do, I'd be happy to make them as I know Saloon like the back of my hand, but I'd need to be able to make changes to your branch/ this PR.

Sammyjo20:
@aydinfatih Also

- What about things like timeout and request options? Could we use the `config` in Saloon and merge that into the SoapConfig to allow people to customise the SoapClient?

Sammyjo20:
How do things like request body work with this? It's been a while since I've used the `SoapClient` 😅 I know you can pass in SoapHeaders but what if someone has a custom body they need to send?

aydinfatih:
Hi @Sammyjo20  
First of all, sorry for my late reply. 
You are welcome. I hope it will be useful to someone. Anyone can send a commit to PR. The important thing is that this work comes out in the best way, we can try to add the feature in the best way together. 
 I tried to answer your questions to the best of my knowledge. 

> Could you write a test for sending a request synchronously / asynchronously using this sender?
- For `SoapClient`, I have only tried sync before, I don't know enough about async request with `Soap`. As far as I searched, I could not find an example as async.
> Could you also write some tests to ensure the right endpoints are being built?
- I will do it.
> I think it's cool that you're using the headers in Saloon to create the `SoapHeaders`, but what about HTTP headers? What if we need to send a custom HTTP header? Can this be done with the `SoapClient`, and I wonder how we could do it in Saloon? I think a good solution could be that if the value provided inside of the headers array isn't a `SoapHeader`, then we treat it like a HTTP header.
- To define http headers for a `SoapClient`, you need to set the `stream_context` value in the array to the options parameter, which is the second parameter of the `SoapClient`. So soap headers and http headers are defined differently. But I think we can accept SoapHeaders as follows instead of creating them ourselves. If the array comes, we can create the `SoapHeader` as in PR. What is your opinion about this?

```PHP
protected function defaultHeaders(): array
{
    return [
        new SoapHeader(namespace:'', name: 'Action', data: 'http://tempuri.org/IWCFService/Test')
    ];
}
```
> I also think perhaps we need a new abstract request class called SoapRequest or something like that where we alias the resolveEndpoint method to resolveAction or something similar? As that is the actual action that is being invoked from the WSDL file in the base URL?
- Yes, maybe there could be a SoapRequest here. I stuck to the existing `Request` class in order not to change user habits too much. Yes, right `resolveEndpoint` corresponds to the method that should be invoked in WSDL.
> How come you have renamed Response to HttpResponse? Additionally, the changes to the interface would result in a breaking API change, so please could you leave any breaking changes to the interfaces. I'm happy if we just have one Response class and not separate SoapResponse and HttpResponse. This is because as explained below, we might be able to create a PSR response from the SoapResponse.
- In this regard, Soap responses actually have different values than Guzzle responses. The methods `fromPsrResponse`, `getPsrResponse`, `stream` and `close` would be empty in Soap responses. For this reason I created a separate contract and proceeded in this way. Also `SoapResponse` must accept a `SoapClient` parameter. `HttpResponse` must accept a value from the `ResponseInterface` class.  There are also differences in the `__construct` method parameters. I can rename `HttpResponse` to `Response` and the `Response` class to `BaseResponse`. This way there will be no change to the existing api. 
> I also wonder if we can perhaps create a PSR response from the SoapClient's response, so we don't have to worry about removing the close capabilities and things like that. You can use Guzzle's Utils class. It has a useful Utils::streamFor() method where you can create a stream interface from a string.
- I actually did some experiments on this subject, but I couldn't convert `SoapClient` response to PSR response. You can also experiment on this subject. I don't think I have as much experience with the `Guzzle` library as you do.
> What about things like timeout and request options? Could we use the config in Saloon and merge that into the SoapConfig to allow people to customise the SoapClient?
- Yes, here I passed the config values for the options value, which is the 2nd parameter of SoapClient. People can customize SoapClient or add timeout in defaultConfig.
```PHP
$this->client ??= new SoapClient(
    wsdl: $pendingRequest->getConnector()->resolveBaseUrl(),
    options: $pendingRequest->config()->all()
);
```
> How do things like request body work with this? It's been a while since I've used the SoapClient 😅 I know you can pass in SoapHeaders but what if someone has a custom body they need to send?
- Body parameters can be passed in the `defaultQuery` method.  I don't know about the custom request body. I reviewed through the Soap libraries but I couldn't find a custom body example.

Sammyjo20:
Hey @aydinfatih, many thanks for getting back to me, no stress on replying fast. I'm actually taking a small break from open source work for the next few weeks, so I may be a little slow to respond and work on this PR, but I'd certainly love to work on it and like you said, come up with something that feels native to Saloon and we are all happy to maintain. Here's some answers to your questions too and some clarifications on mine :D 

### Soap/HTTP Headers

> We can accept SoapHeaders as follows instead of creating them ourselves. If the array comes, we can create the SoapHeader as in PR. What is your opinion about this?

Do you mean checking if each value is an instance of a `SoapHeader`? and If it is an instance of a SoapHeader, put it into the SoapClient's headers, but if the value is anything else (e.g a string) then pass those values into the stream options? If that's the case then that was I was thinking too!

### Abstract Classes

Regarding the abstract class, great 👍 it could just extend the base `Request` class with the `resolveEndpoint` method renamed to something more soap-like. 

### Responses

> In this regard, Soap responses actually have different values than Guzzle responses. The methods fromPsrResponse, getPsrResponse, stream and close would be empty in Soap responses. For this reason I created a separate contract and proceeded in this way. Also SoapResponse must accept a SoapClient parameter. HttpResponse must accept a value from the ResponseInterface class. There are also differences in the __construct method parameters. I can rename HttpResponse to Response and the Response class to BaseResponse. This way there will be no change to the existing api.

I understand that the SoapResponse is very different to a Guzzle/PSR response. Recently I have spent a lot of time with Guzzle's responses as well as the various HTTP-related PSRs like PSR-7 for messages (which is what Guzzle uses under the hood). I think we could use Guzzle's PSR-17 factory classes to create a PSR response from the SoapResponse and standardise it. We can even convert the raw XML into a Stream just like a Guzzle response. I will look into this, so I'll self-assign myself to it.

### Config

> Yes, here I passed the config values for the options value, which is the 2nd parameter of SoapClient. People can customize SoapClient or add timeout in defaultConfig.

Amazing. I think it would be good to make sure that `timeout` and `connect_timeout` is fully supported, and if not we could throw an exception in case someone tries.

### Body / XML Content

> Body parameters can be passed in the defaultQuery method. I don't know about the custom request body. I reviewed through the Soap libraries but I couldn't find a custom body example.

Okay, if we're not going to be supporting the traditional `HasJsonBody`, `HasMultipartBody` etc we should probably throw an exception if the request implements the `HasBody` interface. This way if someone accidentally does this, we will throw an exception. 

This is down to my lack of understanding fully of how SOAP works. I've worked a lot in XML before, and in the past I've just built up the XML manually and passed in additional arguments. How would someone go about sending other properties that aren't headers in Soap? Or do people just use headers to pass data. For example, what if I am submitting a form? Apologies for the lack of understanding, I will also do my research on this! :D

Sammyjo20:
Sorry @aydinfatih I missed the async request point I made. Again this is something I am happy to do myself. I understand that PHP's SoapClient probably doesn't support asynchronous requests, but I think I'd probably just wrap the sending logic around a Guzzle Promise. This way, the promise can be resolved when the application needs it to be and then it fulfils Saloon's "sendAsync" requirement.

Sammyjo20:
Hey @aydinfatih many thanks, once again for proposing this awesome PR for Saloon to introduce the SoapSender into the codebase. I'm currently working on v3 of Saloon which is going to be a much smaller release but one of the main parts of the rebuild is full PSR-7 support in the senders. Unfortunately this PR won't work with that since the SoapClient is its own thing and doesn't support PSR-17, PSR-18 and PSR-7.

I think the best thing moving forward is to make a separate package for the Soap support. I wonder if we can still send the requests with something like Guzzle, but the request body has to be formatted with the traditional soap headers and envelope etc? 

I've had a look around and I've found one library which may be somewhat useful? https://github.com/phpro/soap-client

Sammyjo20:
Hey @aydinfatih many thanks again for your help with this PR - for now I'm going to close it, but I am very open for this being a plugin that people could install in the future. Unfortunately I can't think of a way of supporting this PR for v3 since Saloon is moving to a more PSR-generic route, which will prevent the likes of this SoapClient from working. 

Very happy to revisit soon!

Sammyjo20:
Hey @aydinfatih @deligoez I hope that you don't mind me mentioning you here - I've revisited the idea of having SOAP support in Saloon - but not in the form of a separate sender, rather a simpler implementation but still a pretty nice way of defining SOAP requests. I've created a discussion with my idea of a lovely implementation where we could bounce ideas off each other. It would be good to get your thoughts on it! 🤠

Discussion: https://github.com/saloonphp/saloon/discussions/309

Just to make clear - I'm looking for something simple but effective - and not really something that is going to take a super long time to build unless it's worth the time investment - but with it being a separate package it will invite future maintenance and improvement.

PR #252, opened 2023-07-04
Explicitly cast "expires_in" field to an integer

mabdullahsari:
Hi 👋

This PR fixes an issue that might emerge when using `declare(strict_types=1)` in conjunction with an OAuth server that sends string values via the `expires_in` field instead of an integer.

As you may already know, PHP happily accepts values (type juggling) with the wrong type if `strict_types` is turned off, but fatally crashes if it's enabled.

I've added an extra `is_numeric` check just in case.

(FWIW, this happened to me using the `ClientCredentialsGrant`.)

Thanks 🙌


Sammyjo20 (approved):
Thank you! 🙌

Issue #240, opened 2023-06-22
How to add array to Query Params?


derpoho:
Howdy!

i am wondering how i can add an array to the query params?

I got the issue here on the Pinecone Library: https://github.com/probots-io/pinecone-php/issues/3

Basically i need to send HOSTNAME/vectors/fetch?ids=vector-id-1&ids=vector-id-2
But the current implementation is wrong, since i join the ids with comma, which isn't supported by pinecone.
Here is the current relevant part in the code: https://github.com/probots-io/pinecone-php/blob/master/src/Requests/Index/Vectors/Fetch.php#L58

i tried setting the ids as an array, but thats not working either.

Any ideas?

Sammyjo20:
Hey @derpoho 

This is a bit of a tricky one, and a strange implementation of query parameters from Pinecone's API side - I wonder if this is a standard of query parameter arrays that I'm not aware of. I'll have to do some digging.

Currently, Saloon has a `ArrayStore` which expects `keys` and `values` as the query parameters. I wonder if there's a way that the user could set the query parameters (even at a string level) before the request is sent. Within version 3, I will be constructing a PSR-17 URI class based on the URL and query parameters - I wonder if this could be modified for such situations.

Going back to version two though... I wonder if I could implement some sort of DTO as the value, in these very custom situations...

```php
protected function defaultQuery(): array
{
     return [
         'id' => Query::list([1, 2, 3]); // By default will create ?id=1&id=2&id=3
         'name' => Query::list([1, 2, 3], ','); // With the separator it could do ?id=1,2,3
     ];
}
```

I'll have to see if Guzzle will accept this kind of implementation 😅

derpoho:
Thanks for the reply! It's definitely not standard by Pinecone ;) But anyway, for now i have fixed it by applying middleware to guzzle and rewriting the query 🙈 Ugly, but works for now. 

I am not sure if you even should support that case, is it worth the effort? 

After some digging i found some examples for that one, but everybody is bashing the implementation on the API Side ;)

Sammyjo20:
Hmm I personally can't see any place where this kind of logic is implemented 😅 I've certainly not experienced it but I won't put it past some of the APIs that I deal with.

I reckon from version three I could build up the PSR-17 URI and then allow it to be modified, so at least you don't have to go right into Guzzle to modify it - or maybe even provide a `onPsrRequest` method that is invoked to allow people to modify/debug the PSR request **right** before it's sent (from v3) 💡

[...]

PR #257, opened 2023-07-16
Add Query build utility on Guzzle request

yuters:
This is related to issue #240 

I am currently using Saloon with the Gmail API. There's a query param here called `labelIds[]`
https://developers.google.com/gmail/api/reference/rest/v1/users.messages/list

Contrary to the parameter name in the doc, it must not have the brackets at the end of the param name in order to work. However, this is exactly what Guzzle does.

if I send the defaultQuery has:

```php
["labelIds" => ["INBOX", "UNREAD"]]
````

Then Guzzle will automatically create this query string on the request:
```
?labelIds[]=INBOX&labelIds[]=UNREAD
```

With the small change in this pull request, it renders to what Google expects, which is:
```
?labelIds=INBOX&labelIds=UNREAD
```

Although this might be a breaking change initially for some other API clients, if I'd like to still use the previous behaviour, I could just add the [] in my array key name.

### DIFF
--- a/src/Http/Senders/GuzzleSender.php
+++ b/src/Http/Senders/GuzzleSender.php
@@ -7,6 +7,7 @@ namespace Saloon\Http\Senders;
 use Exception;
 use Saloon\Enums\Timeout;
 use GuzzleHttp\HandlerStack;
+use GuzzleHttp\Psr7\Query;
 use GuzzleHttp\Psr7\Request;
 use Saloon\Contracts\Sender;
 use GuzzleHttp\RequestOptions;
@@ -169,7 +170,7 @@ class GuzzleSender implements Sender
         }
 
         if ($pendingRequest->query()->isNotEmpty()) {
-            $requestOptions[RequestOptions::QUERY] = $pendingRequest->query()->all();
+            $requestOptions[RequestOptions::QUERY] = Query::build($pendingRequest->query()->all());
         }
 
         foreach ($pendingRequest->config()->all() as $configVariable => $value) {


Sammyjo20:
Hey @yuters thank you for the PR 🙌

I've done a bit of digging and it looks like what Guzzle does by default is encode using `http_build_query` with PHP_QUERY_RFC3986. By default, the `Query::build()` method also uses PHP_QUERY_RFC3986 by default, but it uses `urlencode` instead of `http_build_query`.  so I wonder if this will be the same effect. 

See:  https://github.com/guzzle/guzzle/blob/7.7/src/Client.php#L422 
And: https://github.com/guzzle/psr7/blob/2.5/src/Query.php

What are your thoughts? Maybe there's a Saloon way I can implement to get around this as a breaking issue, maybe an interface or option that can be provided.

yuters:
I'm really not sure how to implement something like this, but I can give you an example on how I've made this work using the sender middleware. It's a bit hacky but it can work without changing anything in Guzzle or Saloon. Maybe it could give some ideas:

```php
use GuzzleHttp\Psr7\Query;
use Psr\Http\Message\RequestInterface;

class ApiConnector extends Connector
{
    ...

    public function __construct()
    {
        $this->sender()->addMiddleware(function (callable $handler) {
            return function (RequestInterface $request, array $options) use ($handler) {
                $uri = $request->getUri();
                $query = Query::parse($uri->getQuery());
                $keys = array_keys($query);
                if ($results = preg_grep('/\[(.+?)\]/', $keys)) {
                    foreach ($results as $key) {
                        $keyName = preg_replace('/\[(.+?)\]/', '', $key);
                        $query[$keyName][] = $query[$key];
                        unset($query[$key]);
                    }
                }
                $uri = $uri->withQuery(Query::build($query));
                $request = $request->withUri($uri);
                return $handler($request, $options);
            };
        });
    }

    ...
}

Sammyjo20:
Hey @yuters many thanks for this - although since this is a breaking change, I'm unable to accept this PR in v2 - although I'd be more than happy to revisit this change for v3 which I am currently working on.

PR #269, opened 2023-08-09
Feature | Show request endpoint in MockClient exception message

Joel-Jensen:
PR for my feature request in https://github.com/Sammyjo20/Saloon/discussions/219#discussioncomment-6616142 

Displays the missing endpoint in the exception for easier debugging tests.

How Laravel, `Http::preventStrayRequest()` does it:
image

**This PR:**
image

**Before this PR:**
image

### DIFF
--- a/src/Http/Faking/MockClient.php
+++ b/src/Http/Faking/MockClient.php
@@ -159,7 +159,7 @@ class MockClient implements MockClientContract
         }
 
         if (empty($this->sequenceResponses)) {
-            throw new NoMockResponseFoundException;
+            throw new NoMockResponseFoundException($request);
         }
 
         return $this->mockResponseValue($this->getNextFromSequence(), $pendingRequest);
--- a/src/Exceptions/NoMockResponseFoundException.php
+++ b/src/Exceptions/NoMockResponseFoundException.php
@@ -4,10 +4,12 @@ declare(strict_types=1);
 
 namespace Saloon\Exceptions;
 
+use Saloon\Contracts\Request;
+
 class NoMockResponseFoundException extends SaloonException
 {
-    public function __construct()
+    public function __construct(Request $request)
     {
-        parent::__construct('Saloon was unable to guess a mock response for your request, consider using a wildcard url mock or a connector mock.');
+        parent::__construct("Saloon was unable to guess a mock response for your request [{$request->resolveEndpoint()}], consider using a wildcard url mock or a connector mock.");
     }
 }


Sammyjo20 (changes requested):
Looks great, thank you @Joel-Jensen. Would you mind checking over PHP-Stan as well as update any tests to include this? Thank you!

juse-less:
@Joel-Jensen The only thing you should need to do is rebase from latest v3, where [Sam recently downgraded the PHP constraint (8.2 -> 8.1) in composer.json](https://github.com/saloonphp/saloon/commit/3736b8edb2dcfc4393cf2ac0b9e15a4fa7e4c72a#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34L22).
Then PHPStan should stop complaining.

I can't see anything that PHPStan should complain about in your actual code changes. 👍

Joel-Jensen:
Seems like that branch isn't that stable right now, should I just merge this to v2?

Sammyjo20:
@Joel-Jensen I can get the tests passing for v3 - I'll sort this a little bit later. Happy to leave PHPStan for now as v3 isn't ready yet

Sammyjo20:
If you catch up with master now, it should pass 👍

src/Http/Faking/MockClient.php
```diff
@@ -159,7 +159,7 @@ public function guessNextResponse(PendingRequest $pendingRequest): MockResponse|
         }
 
         if (empty($this->sequenceResponses)) {
-            throw new NoMockResponseFoundException;
+            throw new NoMockResponseFoundException($request);
```
Sammyjo20, line 162: I'm just wondering - would you want the domain + path in the debug log? If you pass the PendingRequest through here, you can do `$pendingRequest->getUrl()` I believe which should return something like: `https://some-api.com/api/path`.

Sammyjo20 (changes requested):

src/Http/Faking/MockClient.php
Joel-Jensen: I would say that it depends on how Saloon::fake([]) will work in v3, thinking regarding my feedback https://github.com/saloonphp/saloon/discussions/219#discussioncomment-6616142. If Saloon::fake() will continue to only expect `/api/path` then I would prefer to get that. But if it will work to send full URL to Saloon::fake() in the future then yeah full URL would be more nice.

Sammyjo20:
Thanks @Joel-Jensen !

PR #276, opened 2023-08-26
Feature | V3 - Allow Middleware To Be Reordered

Sammyjo20:
This PR removes the "prepend" method on the middleware and replaces it with a new argument "order". This accepts a `PipeOrder` class where you can define if a middleware should run at the beginning or at the end of the middleware pipeline. If multiple middleware request to run at the start/end it will be in order or who requested first.

This allows us to have middleware that always runs at the end of a request lifecycle for things like debugging and at the beginning of the response lifecycle for events/debugging and to run before the "AlwaysThrowOnErrors" trait.

This PR also changes the middleware order.

### DIFF
--- a/src/Contracts/MiddlewarePipeline.php
+++ b/src/Contracts/MiddlewarePipeline.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Saloon\Contracts;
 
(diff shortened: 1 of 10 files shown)


Sammyjo20 (commented):
Self review

PR #281, opened 2023-09-03
Feature | V3 - Global Retry System

Sammyjo20:
This PR adds the ability to customise global "tries" on a connector and request level without having to use the `sendAndRetry` method each time when sending requests. This is a great feature to have because if you have an unstable API, you can configure the "tries" at a connector level instead of having to use `sendAndRetry` every time you send a request. This is also useful for SDK developers as they can bake-in retry functionality.

Previously, the `send` method on the connector would simply send the request, and you would need to use the `sendAndRetry` method to use the retry functionality. Now, the `send` method has the retry functionality built into it, but will lay dormant until a number of `tries` has been defined. The `sendAndRetry` method still exists and still can be used on an ad-hoc basis.

To configure the global retry functionality, all you have to do is define the `tries` property on either your connector or request.

```php
class ForgeConnector extends Connector
{
     public ?int $tries = 3;

     // ...
}
```

Then when you send your request like usual, it will be retried up to 3 times before throwing a request exception.

```php
$forgeConnector->send(new GetServersRequest);
```

Behind the scenes, Saloon will use the `throw` method on the connector. This means if you have customized if a request is deemed a failure, this will be used to determine that failure in the retry logic. This is perfect for those APIs that don't provide an "5xx" or "4xx" status code.

You can also configure "retryInterval", "throwOnMaxTries" and overwrite a closure to define a custom handler when retrying. This could be refreshing an access token or simply refusing to retry. 

```php
class ForgeConnector extends Connector
{
     /**
     * The number of tries a request will be attempted
     *
     * Set to null to disable the retry functionality.
     */
    public ?int $tries = null;

    /**
     * The interval between attempting to retry a request
     *
     * Set to null to disable the interval.
     */
    public ?int $retryInterval = null;

    /**
     * Should Saloon throw an exception on max tries?
     *
     * Set to null to always throw after maximum tries.
     */
    public ?bool $throwOnMaxTries = null;

    /**
     * Handle when the request is about to be retried.
     */
    public function handleRetry(FatalRequestException|RequestException $exception, Request $request): bool
    {
        return true;
    }
}
```

## Notable file changes

Apologies, this PR also includes a bunch of interface tidy-ups and small changes to other parts of the Saloon codebase. The main files you will be interested in viewing are:

- The `SendsRequests` trait
- The `HasRetry` trait


Sammyjo20 (commented):
Self review

src/Traits/Connector/SendsRequests.php
```diff
@@ -24,20 +21,87 @@ trait SendsRequests
...
+     * @param callable(\Throwable, \Saloon\Contracts\Request): (bool)|null $handleRetry
      * @throws \ReflectionException
      * @throws \Throwable
      */
-    public function send(Request $request, MockClient $mockClient = null): Response
+    public function send(Request $request, MockClient $mockClient = null, callable $handleRetry = null): Response
     {
-        $pendingRequest = $this->createPendingRequest($request, $mockClient);
+        $maxTries = $request->tries ?? $this->tries ?? 1;
+        $retryInterval = $request->retryInterval ?? $this->retryInterval ?? 0;
```
ash-jc-allen, line 35: Similar to how you've defaulted `$maxTries` to be `1` on line 35 if the value is invalid, do you think it's worth doing the same with the `$retryInterval`? You could maybe default it to `0` if negative numbers are passed in?

From having a quick tinker, it looks like `usleep` (used on line 51) will throw an exception if you pass a negative number to it. Maybe that's the desired behaviour that you want?

But I just wondered whether you'd want to keep it consistent so that if either the `$maxTries` or `$retryInterval` are invalid, they'll be both to reset to default values? Or maybe they should both throw exceptions to highlight to the dev that they've not set something up right in their connectors or requests? At the moment, I think those two fields handle invalid values differently from each other.

This might just be a bit nitpicky though, so feel free to disregard it! 😄

src/Traits/HasTries.php
```diff
@@ -0,0 +1,41 @@
...
+     * Set to null to disable the retry functionality.
+     */
+    public ?int $tries = null;
+
+    /**
+     * The interval between attempting to retry a request
+     *
+     * Set to null to disable the interval.
+     */
+    public ?int $retryInterval = null;
```
ash-jc-allen, line 27: This is only a personal preference, but I always find it really helpful when the property name or doc block says the unit that the interval/time should be in (seconds, milliseconds, microseconds, etc). I guess it's not too hard to dive into the code and figure it out, but it's nice to see it at a quick glance because I'm lazy haha!

src/Traits/HasTries.php
```diff
@@ -0,0 +1,41 @@
...
+     * Set to null to always throw after maximum tries.
+     */
+    public ?bool $throwOnMaxTries = null;
+
+    /**
+     * Handle when the request is about to be retried.
+     */
+    public function handleRetry(FatalRequestException|RequestException $exception, Request $request): bool
+    {
+        return true;
```
ash-jc-allen, line 46: Imagine I wanted to only retry my requests if the server returned a `5xx` error, and I don't want to retry any `4xx` errors. Am I right in saying this is where I'd write this logic? 🙂

ash-jc-allen (commented):
Hey @Sammyjo20! I've not run any of this code, but I've been reading through it to spot any glaringly obvious errors.

I think it looks great! There are a couple of small comments I've made, but I think they're mainly personal preferences, so feel free to ignore them.

I look forward to using this when it's released! 😄

src/Traits/Connector/SendsRequests.php
Sammyjo20: This is a good point! Yep I'll add a minimum in here as well. You can't sleep from a negative number!

src/Traits/HasTries.php
Sammyjo20: Good point! I've given the docblocks a new lick of paint 🙌

Screenshot 2023-09-05 at 19 41 55

src/Traits/HasTries.php
Sammyjo20: Yes that's right! You can do it if the exception was a `RequestException`.

```php
if (! $exception instanceof RequestException) {
     return;
}

$response = $exception->getResponse();

return $response->status() === 400;
```

PR #282, opened 2023-09-03
Feature | V2 Prevent Stray Requests

Sammyjo20:
This PR introduces a new option you can use in your tests to prevent real API calls from happening in your tests. 

```php
Config::preventStrayRequests();
```


Sammyjo20 (commented):
Self review

PR #286, opened 2023-09-08
Feature | V3 Remove Interfaces

Sammyjo20:
This PR removes lots of internal Saloon interfaces that made things more difficult to maintain. Without these interfaces, v3 will be able to have many new features shipped as minor patches! 🔥


Sammyjo20 (commented):
Self review

PR #296, opened 2023-09-19
Fix | V3 - Improve middleware order and add support for PHP 8.3

Sammyjo20:
This PR introduces some final changes to how the `PendingRequest` is constructed. While working on v3, I attempted to migrate every action in the`PendingRequest` into a middleware. When working on my Saloon plugins like the cache plugin, I realised that by running everything in middleware, there was a chance that middleware could be registered before things like headers, config and other middleware were merged into the `PendingRequest` - and things I wouldn't want to be overwritten could be changed. 

I have mulled over and revised this order many times, but I have finally managed to crack it.

When you send a request a, `PendingRequest` class is created which is used as an intermediary class between your connector and requests. It's a place where everything can be merged into one class and everything can be standardized. The order in which things are merged is optimized for the best developer experience. 

## The New Middleware / Execution Order
The new order of execution is as follows:

1. Register Global Middleware
2. Register Mock Middleware
3. Boot Plugins
4. Merge Request Properties
5. Merge Body
6. Merge Delay
7. Run authenticators (and make any future ones run instantly)
8. Boot Connector / Request
9. Delay Middleware
10. Run Middleware Pipeline

When the middleware pipeline is executed, the middleware will be executed in the following order:

1. Global Middleware
2. Mock Middleware
3. Plugin Middleware
4. User Middleware
5. Boot Middleware
6. Delay Middleware

### Global middleware and mock middleware is now executed first
Previously, the global middleware and mock middleware would run after the user's middleware, potentially overwriting anything the user has set. This was initially written because I assumed the user may want to "set" the mock client inside of middleware. Now that I realise that's not really a use case I have moved it to run first. This has a couple of benefits:

- The mock response is determined straight away, so all future middleware can see the "fake response" that is going to be sent. This is great for debugging.
- The fake response can be easily swapped out with something different by using `$pendingRequest->setFakeResponse()` which gives the user better control.

### Authenticators are now executed before middleware starts
Previously, you wouldn't be able to see the result of `$request->withTokenAuth()` or `$request->authenticate()` until you `dd()`ed the PendingRequest's headers. This isn't very developer friendly as it made it appear that the headers were never being set. Now, we will apply the authenticators before middleware is authenticated. This means that inside of middleware, you will be able to see any authentication tokens applied and also choose to overwrite them if needed. This is also much better for debugging.

## Reviewing

I recommend opening the `PendingRequest` file and reading through it - please let me know if there's anything in here which could be improved for better readability/performance.

https://github.com/saloonphp/saloon/blob/942d9740cf8ae626d6def7e8919170f4b6e86244/src/Http/PendingRequest.php

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ ubuntu-latest, windows-latest ]
-        php: [ 8.1, 8.2 ]
+        php: [ 8.1, 8.2, 8.3 ]
         stability: [ prefer-lowest, prefer-stable ]
 
     name: P${{ matrix.php }} - ${{ matrix.stability }} - ${{ matrix.os }}
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     "homepage": "https://github.com/saloonphp/saloon",
     "require": {
         "php": "^8.1",
-        "guzzlehttp/guzzle": "^7.7",
+        "guzzlehttp/guzzle": "^7.6",
         "guzzlehttp/promises": "^1.5 || ^2.0",
         "guzzlehttp/psr7": "^2.0",
         "psr/http-factory": "^1.0",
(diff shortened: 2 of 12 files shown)


Sammyjo20 (commented):
Self review

PR #298, opened 2023-09-20
Feature | V3 - Add `all` method back onto body

Sammyjo20:
This PR must be merged after #296 as it contains changes from that branch.

This PR simply adds back the `all()` method which was previously removed. To reduce headaches and improve backwards compatibility - I have left the method in.


Sammyjo20 (commented):
Self review

PR #305, opened 2023-10-02
feat: remove `SoloRequest`

innocenzi:
This pull request is a suggestion at removing `SoloRequest`. Honestly, I don't think it's useful to make developers have to differentiate `Request` and `SoloRequest`.

The two elements that made a `SoloRequest` work were moved to `Request`, and `SoloRequest` was marked as deprecated.

I'm willing to PR the docs if this is accepted.


Sammyjo20:
Hey @innocenzi many thanks for your PR and your suggestion 🤠

I am however going to have to respectfully decline this proposal. I've heard from numerous people that the `SoloRequest` is a very useful class for people to have, and I've found it useful myself personally too. You're free to carry on using regular requests without having to use the `SoloRequest` class in your application.

Additionally, I wouldn't be able to merge this PR anyway as it would involve a breaking change. ☺️

Have a great day!

innocenzi:
Hey @Sammyjo20, thanks for the answer!

I made sure that the pull request is not a breaking change—I use `SoloRequest` myself more than regular connector requests!

The intent of the PR is to avoid having to make the developer think what base `Request` class they should extend, whether a normal request or a request associated with a connector is needed. I also always thought that the term "solo" request was an odd word for something that's just a regular request. 😄 

Could you perhaps take a second look at what the pull request does? It basically merges `SoloRequest` into `Request`, but keeps `SoloRequest` accessible to avoid the breaking change. Let me know if you need more clarification.

Sammyjo20:
Hey @innocenzi appreciate the clarification of the PR. I've taken a second look, and I've decided that I would like to keep the `SoloRequest` in Saloon v3. 

The `SoloRequest` class exists because the regular `Request` class does not contain the `send` or `sendAsync` methods. This is a deliberate design choice because for larger API integrations I would like people to use the `Connector` class to send requests - like the following:

```php
$connector->send(new MyRequest);
```

Adding the `HasConnector` trait on the `Request` class means that the `send` and `sendAsync` methods would be available for every single request, which I'm not a fan of because that goes against Saloon v2+'s design choice and makes it difficult for the developer to choose which way they should send a request. Additionally, someone might get confused and accidentally call `send` on a request that doesn't contain a full request URL, resulting in an error. This is why I decided to make the `SoloRequest`. As explained in the documentation:

> While Saloon's typical setup of a connector and requests is great, sometimes all you need is to make a single request to a service. For scenarios like these, you may create a "SoloRequest" instead of making a connector and a single request. This saves you from having to create additional classes.

The `SoloRequest` uses the `HasConnector` class because when using a `SoloRequest` you aren't likely going to have a `Connector` class. This is why the `send` and `sendAsync` methods exist only on the `SoloRequest` class.

```php
(new MySoloRequest)->send();
```

All the best
Sam

innocenzi:
While I don't agree with the design choice, I understand and respect the reasoning 👍 

Thanks for taking a second look!

PR #312, opened 2023-10-12
Fix | V3 - Using Just Connector Body

### LINKED ISSUE #311: Error when only connector has a body and request does not
I have a situation with a Connector that implements the `HasBody`-interface and a Request that doesn't. Now I'm getting an `__clone method called on non-object`-error.

I think it's because of the following code: https://github.com/saloonphp/saloon/blob/dc0284d6dcbe21ee7f4e0fe746a75c7431e7fba5/src/Http/PendingRequest/MergeBody.php#L43

It tries to clone `$requestBody` while it's `null` and then fails, which is strange because of the null coalescing operator (`??`).

When I implement the `HasBody`-interface to my Request too, the error disappears.

_PHP version: 8.2.7_

Sammyjo20:
This PR fixes an issue where if you had body on just the connector, Saloon would throw an exception because it was trying to call `clone` on a the `$requestBody` variable which has the value `null`. You cannot clone a non-object variable.

Fixes #311

### DIFF
--- a/src/Http/PendingRequest/MergeBody.php
+++ b/src/Http/PendingRequest/MergeBody.php
@@ -22,8 +22,8 @@ class MergeBody
         $connector = $pendingRequest->getConnector();
         $request = $pendingRequest->getRequest();
 
-        $connectorBody = $connector instanceof HasBody ? $connector->body() : null;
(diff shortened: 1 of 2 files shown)


Sammyjo20 (commented):
Self review

PR #314, opened 2023-10-31
Fix | `hasRequestFailed` method only works to mark successful requests as failures

gdebrauwer:
In a project, I am using an API call that returns a list of items. If there are no items, then the API call (sadly) returns a 404 response. That means I don't want an exception in that case, but just an empty collection. In my Saloon request, I implemented the `hasRequestFailed` method and I ensured it returns `false` when the response has status 404:
```php
class SomeRequest extends Request
{
    public function hasRequestFailed(Response $response): ?bool
    {
        return ($response->serverError() || $response->clientError()) && $response->status() !== 404;
    }
}
```
When I test this code, Saloon still throws an exception when I get a 404 response. That is because the current implementation of the `hasRequestFailed` method only allows you to mark successful requests as failures, but not the other way around. This PR fixes that.

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -336,10 +336,11 @@ class Response
     {
         $pendingRequest = $this->getPendingRequest();
 
-        $hasRequestFailed = $pendingRequest->getRequest()->hasRequestFailed($this) || $pendingRequest->getConnector()->hasRequestFailed($this);
+        $hasRequestFailedAccordingToRequest = $pendingRequest->getRequest()->hasRequestFailed($this);
+        $hasRequestFailedAccordingToConnector = $pendingRequest->getConnector()->hasRequestFailed($this);
 
-        if ($hasRequestFailed === true) {
-            return true;
+        if ($hasRequestFailedAccordingToRequest !== null || $hasRequestFailedAccordingToConnector !== null) {
+            return $hasRequestFailedAccordingToRequest || $hasRequestFailedAccordingToConnector;
         }
 
         return $this->serverError() || $this->clientError();
--- a/src/Http/Senders/GuzzleSender.php
+++ b/src/Http/Senders/GuzzleSender.php
@@ -170,9 +170,10 @@ class GuzzleSender implements Sender
 
                     $response = $this->createResponse($guzzleResponse, $pendingRequest, $psrRequest, $guzzleException);
 
-                    // Throw the exception our way
+                    // Throw the exception our way if there is an exception.
+                    // Otherwise we'll return the response.
 
-                    throw $response->toException();
+                    return ($exception = $response->toException()) ? throw $exception : $response;
                 }
             );
     }
--- a/src/Traits/Connector/SendsRequests.php
+++ b/src/Traits/Connector/SendsRequests.php
@@ -74,15 +74,7 @@ trait SendsRequests
 
                 $response = $pendingRequest->executeResponsePipeline($response);
 
-                // We'll check if our tries is greater than one. If it is, then we will
-                // force an exception to be thrown if the request was unsuccessful.
-                // This will then force our catch handler to retry the request.
-
-                if ($maxTries > 1) {
-                    $response->throw();
-                }
-
-                return $response;
+                return $response->throw();
             } catch (FatalRequestException|RequestException $exception) {
                 // We'll attempt to get the response from the exception. We'll only be able
                 // to do this if the exception was a "RequestException".
--- a/tests/Feature/PoolTest.php
+++ b/tests/Feature/PoolTest.php
@@ -91,17 +91,25 @@ test('if a pool has a request that cannot connect it will be caught in the handl
 });
 
 test('if a pool has a failed response that should be treated as a successful response', function () {
-    $count = 0;
+    $responseCount = 0;
+    $exceptionCount = 0;
 
     $pool = (new TestConnector)->pool([
         new ErrorRequestThatShouldBeTreatedAsSuccessful,
     ]);
 
-    $pool->withResponseHandler(function (Response $response) use (&$count) {
+    $pool->withResponseHandler(function (Response $response) use (&$responseCount) {
         expect($response)->toBeInstanceOf(Response::class);
         expect($response->status())->toBe(404);
 
-        $count++;
+        $responseCount++;
+    })->withExceptionHandler(function (RequestException $exception) use (&$exceptionCount) {
+        $response = $exception->getResponse();
+
+        expect($response)->toBeInstanceOf(Response::class);
+        expect($response->status())->toBe(404);
+
+        $exceptionCount++;
     });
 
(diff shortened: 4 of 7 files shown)


Sammyjo20:
Hey @gdebrauwer thank you for posting this - apologies I’ve been really busy this week so I’ll try and look next week for you 👍

src/Http/Response.php
Sammyjo20, line 1: Just getting my head around this, so currently because the default value in `hasRequestFailed` is null - this was causing issues with your request?

Because `false || null` = false in your case, so that it was always going to the error below, but with this, if either are not null (e.g have been overwritten) the condition is checked for at least one of them being true

false || null = false
false || true = true
true || true = true

src/Http/Senders/GuzzleSender.php
```diff
@@ -170,9 +170,10 @@ function (TransferException $guzzleException) use ($pendingRequest, $psrRequest)
 
                     $response = $this->createResponse($guzzleResponse, $pendingRequest, $psrRequest, $guzzleException);
 
-                    // Throw the exception our way
+                    // Throw the exception our way if there is an exception.
+                    // Otherwise we'll return the response.
 
-                    throw $response->toException();
+                    return ($exception = $response->toException()) ? throw $exception : $response;
```
Sammyjo20, line 176: I'm a little confused by this one - do you mind explaining a little please?

tests/Feature/PoolTest.php
```diff
@@ -89,6 +90,29 @@
...
+    });
+
+    $promise = $pool->send();
+
+    expect($promise)->toBeInstanceOf(PromiseInterface::class);
+
+    $promise->wait();
+
+    expect($count)->toEqual(1);
+});
```
Sammyjo20: Please could we also add a similar test for a synchronous request?

Sammyjo20 (commented):
Awesome PR, thank you @gdebrauwer - just a couple of questions

Sammyjo20:
Don't worry about the code style workflow

src/Http/Senders/GuzzleSender.php
gdebrauwer: The `$response->toException()` method can return null if the `shouldThrowRequestException()` method returns false. In that case, we can not throw an exception and we have to return the response.

src/Http/Response.php
gdebrauwer: On my request I have a `hasRequestFailed` method that returns `false`, so I expect the package to treat the response as a successful response. But because of the code `if ($hasRequestFailed === true) { return true;` it is only possible to treat a successful response as failed. When `$hasRequestFailed` is `false,` it still checks if the response is a server or client error. That means that if a response is a server error but the you want to treat that as a successful response (= `hasRequestFailed` returns false), then that currently does not work.

src/Http/Senders/GuzzleSender.php
Sammyjo20: Ah gotcha

tests/Feature/PoolTest.php
gdebrauwer: By adding the test for synchronous requests I now discovered that synchronous requests without retry functionality do not respect the `shouldThrowRequestException` method 🤔 (causing some existing tests to fail)

gdebrauwer: the `shouldThrowRequestException` implementation has a similar issue to the `hasRequestFailed` method.

If `shouldThrowRequestException` on the request class returns `false` then an exception will still be thrown because the `shouldThrowRequestException` on the connector class will always return `true` by default if the request fails
```php
// in Response class

public function shouldThrowRequestException(): bool
{
    $pendingRequest = $this->getPendingRequest();

	// If you customize the shouldThrowRequestException method on the request
	// then that will not have a any effect unless you also override the method on
	// the connector class
    return $pendingRequest->getRequest()->shouldThrowRequestException($this)
		|| $pendingRequest->getConnector()->shouldThrowRequestException($this);
}
```

src/Http/Response.php
Sammyjo20: Yep that makes sense

src/Traits/Connector/SendsRequests.php
```diff
@@ -74,15 +74,7 @@ public function send(Request $request, MockClient $mockClient = null, callable $
...
-                // We'll check if our tries is greater than one. If it is, then we will
-                // force an exception to be thrown if the request was unsuccessful.
-                // This will then force our catch handler to retry the request.
-
-                if ($maxTries > 1) {
-                    $response->throw();
-                }
-
-                return $response;
+                return $response->throw();
```
Sammyjo20, line 77: I think this logic needs to remain the same because by default, Saloon does not throw exceptions if the request has failed. This check was in place because if no one has configured the send and retry, than the "max tries" is only 1. If this is the case, we shouldn't throw an exception here.

Sammyjo20 (changes requested):

tests/Feature/PoolTest.php
Sammyjo20: Hmm I've had a look, and it looks like the `shouldThrowRequestException` only returns `bool` rather than the `hasRequestFailed` was `?bool` - so I don't think this needs changing.

I wonder if this is related to the send and retry functionality because in v3 every request runs through the sendAndRetry logic but only if the `maxAttempts` > 1 than this logic will be executed.

gdebrauwer: The maxAttempts > 1 is the problem I think.

gdebrauwer:
I'm currently on vacation until the beginning of December so I will only have time to update the PR in a couple weeks.

src/Traits/Connector/SendsRequests.php
gdebrauwer: By default it does not throw an exception, but if you want to throw an exception (using `shouldThrowRequestException` method), then the current code will not do that.

Sammyjo20:
Howdy @gdebrauwer are you still happy to have a look at this PR when you get a chance? Thanks.

src/Traits/Connector/SendsRequests.php
gdebrauwer: @Sammyjo20 could you check my response here? ⬆️

gdebrauwer:
> Howdy @gdebrauwer are you still happy to have a look at this PR when you get a chance? Thanks.

I added a comment to your PR review question.

Sammyjo20:
Sorry for the delay @gdebrauwer I have blocked out Sunday to go through these things so I will check again then.

Sammyjo20:
I have made a PR #371 which should fix everything proposed - the only thing I didn't implement is the change in the way responses are thrown. I think you may have misunderstood how `shouldThrowRequestException` works. That method is only used if you call the `$response->throw()` method or if you are using `AlwaysThrowOnError`.

PR #315, opened 2023-10-31
Bugfix: empty response body to json should not fail

gdebrauwer:
Similar to Laravel's Http Client Response, the `json()` method on a response should not throw an exception when the body is empty.


Sammyjo20 (approved):

Sammyjo20:
Thank you!

PR #321, opened 2023-11-03
Feature | Add header authenticator class

markwalet:
I have to consume some weird API's in one of my projects at work. That's why I figured a more general header authenticator class could be of good use. I'm using this one as a custom authenticator in my project right now but I wanted to share it in the case someone else could use it as well.

I couldn't find any dedicated authenticator tests, so I only added the helper method in the `AuthenticatesRequests` trait to keep it consistent with the rest. If I'm missing anything, please let me know. Feel free to close the MR if this kind of authenticator is not desirable in Saloon.

### DIFF
--- a/src/Http/Auth/HeaderAuthenticator.php
+++ b/src/Http/Auth/HeaderAuthenticator.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Saloon\Http\Auth;
+
+use Saloon\Http\PendingRequest;
+use Saloon\Contracts\Authenticator;
+
+class HeaderAuthenticator implements Authenticator
+{
+    /**
+     * Constructor
+     */
+    public function __construct(
+        public string $accessToken,
+        public string $headerName = 'Authorization',
+    ) {
+        //
+    }
+
+    /**
+     * Apply the authentication to the request.
+     */
+    public function set(PendingRequest $pendingRequest): void
+    {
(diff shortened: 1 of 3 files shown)


markwalet:
I tried to reuse this class in other authenticators as well. But I found that it didn't really help with readability:

```php
class TokenAuthenticator extends HeaderAuthenticator implements Authenticator
{

    public function __construct(
        public string $token,
        public string $prefix = 'Bearer'
    ) {
        parent::__construct(
            trim($this->prefix . ' ' . $this->token),
        );
    }
}
```

Sammyjo20:
Hey @markwalet thank you, this looks like a nice PR! 

I was wondering, is this similar to the `TokenHeader` already provided? I wonder if we could customise that authenticator to have a third `$header = 'Authorization'` instead maybe?

markwalet:
I mean that's fine by me as well, but if the prefix becomes optional, there's an extra if-statement in that class to remove the extra whitespace when needed. I'm fine with that but you have the chance that it obfuscates the main use-case for the class. Separating them out into multiple classes makes more sense to me personally. Just to give the end-user some more explicit cases to choose from.

But I'm not the maintainer, either way is fine for me. Just let me know what your preference is :)

Sammyjo20 (changes requested):
I'm happy with the code - would you mind adding a test for this please?

Sammyjo20 (approved):
Awesome PR, thank you!!

PR #326, opened 2023-11-12
Fix | MockClient assertions with fixtures

### LINKED ISSUE #322: Asserting sent request fails on first test when recording responses
Thanks so much for such a great package - it's amazing seeing the effort you put into it!

I've encountered a problem with fixtures which feels like a bug, but may also be intentional. I have created a minimal example of the test I'm trying to run:

```php
use Package\SomeConnector;
use Package\SomeRequest;
use Saloon\Http\Faking\MockResponse;
use Saloon\Laravel\Facades\Saloon;

it('asserts request was sent', function (): void {
    Saloon::fake([
        '*' => fn (PendingRequest $pendingRequest) => MockResponse::fixture(/** name logic  */),
    ]);

    (new SomeConnector())->send(new SomeRequest());

    Saloon::assertSent(SomeRequest::class);
});
```

What I would expect is that this test passes every time, even the very first time when it is recording the response and creating the fixture. Instead, the assertion fails on the first run. The fixture gets created though, and the test then passes on all subsequent runs.

It may be useful to include that I'm using a connector and request from an external SDK package I created.

Is this expected behaviour?

Sammyjo20:
This PR fixes an issue with the `MockClient` assertions when a fixture doesn't exist. Methods on the `MockClient` like `assertSent` would use an array of `recordedResponses` which were previously only recorded after creating a mock-response. When using fixtures, because the first request that was sent was not a `MockResponse`, it was never being recorded. This PR changes Saloon so fixtures are recorded to the mock client.

This should resolve #322.

### DIFF
--- a/src/Http/Middleware/DetermineMockResponse.php
+++ b/src/Http/Middleware/DetermineMockResponse.php
@@ -53,7 +53,7 @@ class DetermineMockResponse implements RequestMiddleware
         // middleware on the response to record the response.
 
         if (is_null($mockResponse) && $mockObject instanceof Fixture) {
-            $pendingRequest->middleware()->onResponse(new RecordFixture($mockObject), 'recordFixture', PipeOrder::FIRST);
+            $pendingRequest->middleware()->onResponse(new RecordFixture($mockObject, $mockClient), 'recordFixture', PipeOrder::FIRST);
         }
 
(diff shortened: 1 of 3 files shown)


Sammyjo20 (commented):
Self review

PR #328, opened 2023-11-16
Fix | Merge the headers from the Connector into the `PendingRequest` when mocking

pascalbaljet:
Hey again! 👋 This PR provides a similar fix as https://github.com/saloonphp/laravel-http-sender/pull/14.

When working with mocks, the headers from the connector are not merged into the `PendingRequest` instance. We are using the `createPsrRequest()` method on this instance to do some logging and saw the headers missing. This PR fixes that.

### DIFF
--- a/src/Traits/Connector/ManagesFakeResponses.php
+++ b/src/Traits/Connector/ManagesFakeResponses.php
@@ -6,6 +6,7 @@ namespace Saloon\Traits\Connector;
 
 use Throwable;
 use Saloon\Http\Response;
+use GuzzleHttp\RequestOptions;
 use Saloon\Http\PendingRequest;
 use Saloon\Contracts\FakeResponse;
 use Saloon\Http\Faking\MockResponse;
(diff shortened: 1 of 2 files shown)


Sammyjo20:
Hey @pascalbaljet I was wondering, why are you using the "headers" option in the config and not the native "headers" API that Saloon provides? Is there something I can improve with that which wouldn't require a change to this code? I'd rather not write Guzzle-specific code in places like this because it moves us further away from being sender agnostic.

onlime:
Hey @Sammyjo20 - Happy New Year!
This is not Guzzle-specific. The only mistake in this PR is to use `GuzzleHttp\RequestOptions::HEADERS` constant instead of a simple `'headers'` string. The Connector's headers (from `defaultHeaders()` in `HasHeaders` trait) should still be merged into the `PendingRequest`. @pascalbaljet Can you please update this PR to no longer depend on `GuzzleHttp\RequestOptions::HEADERS` const?

Or does this only affect `headers` that are set via the Connector's `defaultConfig()`? I still think they should be merged as well like proposed in this PR, as Guzzle (or any other potential sender) respects `headers` in the config.

Sammyjo20:
Hey @onlime - happy new year. Guzzle is the only sender that uses the `headers` configuration option. Are you able to explain how you were using Saloon to require this change @pascalbaljet? As I said before, you should use Saloon's own header API instead of tapping into Guzzle's config options if possible.

Sammyjo20:
Hey @pascalbaljet @onlime what are your thoughts?

PR #329, opened 2023-11-16
Feature | Adding `withRequestMocks()` helper + introducing `HasFakeData` interface

pascalbaljet:
This PR consists of two new features. I worked on it with @onlime, credits to him for the idea!

**Request Mocks**

The first feature is a `withRequestMocks()` method that you may call, for example, on a `Connector`. You may use both Request class names and URLs. Under the hood, it transforms the array into an array of mock responses that are passed to the `MockClient`. Also, you may now pass arrays and strings in addition to using `MockResponse`/`Fixture` instances and callbacks.

```php
ForgeConnector::make()->withRequestMocks([
    GetUserRequest::class => 'Sam',
    GetServersRequest::class => ['Server A', 'Server B'],
    OtherServiceRequest::class => MockResponse::fixture('other-service'),
    WildcardServiceRequest::class => fn (PendingRequest $pendingRequest) => MockResponse::make(['...'])
    
    // or...

    '/user/1' => 'Sam',
    '/servers' => ['Server A', 'Server B'],
]);
```

If you don't provide a response, it simply uses an empty `MockResponse`:

**`HasFakeData` interface**

The second feature is a `HasFakeData` interface that you may use on a Request class. Also here, you may return a string, an array, or a `MockResponse` or `Fixture` instance:

```php
use Saloon\Enums\Method;
use Saloon\Http\Request;
use Saloon\Contracts\HasFakeData;

class GetUserRequest extends Request implements HasFakeData
{
    /**
     * Define the HTTP method.
     */

    /**
     * Define the endpoint for the request.
     */
    {
        return '/user';
    }

    /**
     * Get the fake data for the mocked request.
     */
    public function getFakeData(PendingRequest $request): mixed
    {
        return ['Sam'];
    }
}
```

Now when you pass the request to the `withRequestMocks` method, it will use the `getFakeData` method to mock the response:

```php
ForgeConnector::make()->withRequestMocks([
    GetUserRequest::class,
]);
```

You may even mix requests with and without this interface, as well as URLs:

```php
ForgeConnector::make()->withRequestMocks([
    GetUserRequest::class,
    GetServersRequest::class => ['Server A', 'Server B'],
    '/meta' => ['status' => 'pending'],
]);
```

### DIFF
--- a/src/Contracts/HasFakeData.php
+++ b/src/Contracts/HasFakeData.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Saloon\Contracts;
+
+use Saloon\Http\PendingRequest;
+
+interface HasFakeData
+{
+    /**
+     * Get the fake data for the mocked request.
+     */
+    public function getFakeData(PendingRequest $request): mixed;
+}
--- a/src/Contracts/HasFakeData.php
+++ b/src/Contracts/HasFakeData.php
@@ -4,12 +4,16 @@ declare(strict_types=1);
 
 namespace Saloon\Contracts;
 
+use Saloon\Http\Faking\Fixture;
 use Saloon\Http\PendingRequest;
+use Saloon\Http\Faking\MockResponse;
 
 interface HasFakeData
 {
     /**
      * Get the fake data for the mocked request.
+     *
+     * @return array<mixed,mixed>|string|MockResponse|Fixture
      */
-    public function getFakeData(PendingRequest $request): mixed;
+    public function getFakeData(PendingRequest $request): array|string|MockResponse|Fixture;
 }
--- a/src/Traits/HasMockClient.php
+++ b/src/Traits/HasMockClient.php
@@ -47,6 +47,8 @@ trait HasMockClient
      * Mocks the given requests with the given responses, or with the
      * mocked response from the 'getFakeData()' method if the
      * request has implemented the HasFakeData interface.
+     *
+     * @param array<array-key|class-string, mixed> $requestMocks
      */
     public function withRequestMocks(array $requestMocks): static
     {
@@ -67,9 +69,13 @@ trait HasMockClient
             }];
         })->all();
 
-        return $this->hasMockClient()
-            ? $this->getMockClient()->addResponses($responses)
-            : $this->withMockClient(new MockClient($responses));
+        if ($this->hasMockClient()) {
+            $this->getMockClient()->addResponses($responses);
+
+            return $this;
+        }
+
+        return $this->withMockClient(new MockClient($responses));
     }
 
     /**
--- a/src/Traits/HasMockClient.php
+++ b/src/Traits/HasMockClient.php
@@ -4,7 +4,11 @@ declare(strict_types=1);
 
 namespace Saloon\Traits;
 
+use Saloon\Http\Faking\Fixture;
+use Saloon\Http\PendingRequest;
+use Saloon\Contracts\HasFakeData;
 use Saloon\Http\Faking\MockClient;
+use Saloon\Http\Faking\MockResponse;
 
 trait HasMockClient
 {
@@ -23,6 +27,49 @@ trait HasMockClient
         $this->mockClient = $mockClient;
 
         return $this;
+    }
+
+    /**
+     * Creates a MockResponse instance from the given value if
+     * it is a string or an array. Otherwise, it returns the
+     * value as is.
+     */
+    private function prepareMockResponse(mixed $value): MockResponse|Fixture|callable
+    {
+        if (is_array($value) || is_string($value)) {
+            return new MockResponse($value);
+        }
+
+        return $value;
+    }
+
+    /**
+     * Mocks the given requests with the given responses, or with the
+     * mocked response from the 'getFakeData()' method if the
+     * request has implemented the HasFakeData interface.
+     */
+    public function withRequestMocks(array $requestMocks): static
+    {
+        $responses = collect($requestMocks)->mapWithKeys(function (mixed $value, int|string $key) {
+            // Something like: [UserRequest::class => MockResponse::fixture('user')];
+            // Or URL mocking: ['/user' => ['id' => 123];
+            if (! is_numeric($key)) {
+                return [$key => $this->prepareMockResponse($value)];
+            }
+
+            // Just the Request class name or the URL. Retrieve the fake data from the request.
+            return [$value => function (PendingRequest $pendingRequest) {
+                $request = $pendingRequest->getRequest();
+
+                return $request instanceof HasFakeData
+                    ? $this->prepareMockResponse($request->getFakeData($pendingRequest))
+                    : MockResponse::make();
+            }];
+        })->all();
+
+        return $this->hasMockClient()
+            ? $this->getMockClient()->addResponses($responses)
+            : $this->withMockClient(new MockClient($responses));
     }
 
     /**
--- a/tests/Feature/HasMockClientTest.php
+++ b/tests/Feature/HasMockClientTest.php
@@ -1,0 +1,139 @@
+<?php
+
+declare(strict_types=1);
+
+use Saloon\MockConfig;
+use Saloon\Http\PendingRequest;
+use League\Flysystem\Filesystem;
+use Saloon\Http\Faking\MockResponse;
+use Saloon\Tests\Fixtures\Requests\UserRequest;
+use League\Flysystem\Local\LocalFilesystemAdapter;
+use Saloon\Tests\Fixtures\Connectors\TestConnector;
+use Saloon\Tests\Fixtures\Requests\SoloUserRequest;
+use Saloon\Tests\Fixtures\Requests\UserRequestWithFakeData;
+use Saloon\Tests\Fixtures\Requests\UserRequestWithFakeDataFixture;
+
+$filesystem = new Filesystem(new LocalFilesystemAdapter('tests/Fixtures/Saloon/Testing'));
+
+beforeEach(function () use ($filesystem) {
+    MockConfig::setFixturePath('tests/Fixtures/Saloon/Testing');
+
+    $filesystem->deleteDirectory('/');
+    $filesystem->createDirectory('/');
+});
+
+afterEach(function () {
+    MockConfig::setFixturePath('tests/Fixtures/Saloon');
+});
+
+it('can mock a request with a string', function () {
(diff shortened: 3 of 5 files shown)


src/Traits/HasMockClient.php
```diff
@@ -25,6 +29,49 @@ public function withMockClient(MockClient $mockClient): static
...
+                $request = $pendingRequest->getRequest();
+
+                return $request instanceof HasFakeData
+                    ? $this->prepareMockResponse($request->getFakeData($pendingRequest))
+                    : MockResponse::make();
+            }];
+        })->all();
+
+        return $this->hasMockClient()
+            ? $this->getMockClient()->addResponses($responses)
```
onlime: @pascalbaljet `MockClient::addResponses()` returns `void`, so I propose to change this to:

```php
        return $this->hasMockClient()
            ? tap($this->getMockClient(), fn (MockClient $mockClient) => $mockClient->addResponses($responses))
            : $this->withMockClient(new MockClient($responses));
```

src/Traits/HasMockClient.php
```diff
@@ -25,6 +29,49 @@ public function withMockClient(MockClient $mockClient): static
...
+     * Mocks the given requests with the given responses, or with the
+     * mocked response from the 'getFakeData()' method if the
+     * request has implemented the HasFakeData interface.
+     */
+    public function withRequestMocks(array $requestMocks): static
+    {
+        $responses = collect($requestMocks)->mapWithKeys(function (mixed $value, int|string $key) {
+            // Something like: [UserRequest::class => MockResponse::fixture('user')];
+            // Or URL mocking: ['/user' => ['id' => 123];
+            if (! is_numeric($key)) {
```
onlime, line 58: I would use `is_int()` instead of `is_numeric()` as PHP always converts numeric strings to integers anyway, so even a `'123'` wouldn't be treated as a URL path, but just as a 1-dimensional array index, though assuming we have just passed the request class. Probably irrelevant, though.

src/Contracts/HasFakeData.php
```diff
@@ -0,0 +1,15 @@
...
+namespace Saloon\Contracts;
+
+use Saloon\Http\PendingRequest;
+
+interface HasFakeData
+{
+    /**
+     * Get the fake data for the mocked request.
+     */
+    public function getFakeData(PendingRequest $request): mixed;
```
onlime: maybe replace return type by `array|string|MockResponse|Fixture` - or could this be anything else?

onlime (changes requested):

src/Traits/HasMockClient.php
pascalbaljet: I mainly used `is_numeric` because that's what is used throughout the Laravel codebase: https://github.com/laravel/framework/blob/38ae673608b334c983ba7e729e605eef9729eab8/src/Illuminate/Collections/Arr.php#L829

src/Contracts/HasFakeData.php
pascalbaljet: No that's it I think. It's just a bit long in the implementation:

```php
/**
 * Get the fake data for the mocked request.
 */
public function getFakeData(PendingRequest $request): array|string|MockResponse|Fixture
{
    return ['status' => 'pending'];
}
```

src/Traits/HasMockClient.php
onlime: thx for fixing my fix in 2c744a61c9ea6201b8ce3f05a3f65025a0c87aad 🙂

src/Contracts/HasFakeData.php
onlime: I prefer long explicit return types in interfaces. They can always be narrowed down in the implementation, so the following would be "PHPStan-safe", I guess:

```php
public function getFakeData(PendingRequest $request): array
{
    return ['status' => 'pending'];
}
```

should be possible since PHP 7.4, see [Covariant Returns and Contravariant Parameters](https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters)

having all possible return types in that method also serves as documentation / easier to understand all the sexiness of the magic `withRequestMocks()` that handles it 😎

pascalbaljet: Fixed in e6dd5f9

onlime (approved):

Sammyjo20:
Hey @pascalbaljet and @onlime many thanks for your patience on me looking at this PR! I'm sorry for my slow reply.

Thank you for your efforts in putting this PR and idea together. 

Are you able to explain the use-case/benefits over using the `MockClient` on the connector? I think the idea is cool but I'd like to just know a little bit of the context around the newer way of mocking requests ☺️  I'm personally very happy with how the current `MockClient` works but if I'm missing a specific use-case let me know.

onlime:
Hey @Sammyjo20  – sure, I'm happy to show you my use case where I think it totally makes sense to have some kind of helper/abstraction for Request mocking. We are using mocked/fake responses not just for testing but also for development, and in each integration's constructor, we can enable/disable mocking by configuration. So a very basic example would look like this:

```php
    public function __construct()
    {
        if (! config('acme.api.enable')) {
            $this->withMockClient(new MockClient([
                FetchNote::class  => fn (PendingRequest $request) => MockResponse::make($request->getRequest()->getFakeData()),
                FetchNotes::class => MockResponse::fixture('Acme/notes'),
                CreateNote::class => fn (PendingRequest $request) => MockResponse::make($request->getRequest()->getFakeData()),
            ]));
        }
    }
```

vs. (with this PR merged):

```php
    public function __construct()
    {
        if (! config('acme.api.enable')) {
            $this->withRequestMocks([
                FetchNote::class,
                FetchNotes::class,
                CreateNote::class,
            ]);
        }
    }
```

In the first code snippet you see how blown up it will get, especially since I have just added 3 requests, but in a real SDK/integration, it will be potentially hundreds. We could group them by resource, though, something like `new MockClient([...NoteResource::getRequestMocks(), ...AnotherResource::getRequestMocks()])`, but that does not solve the problem of massively blown up code for the mocked responses.

So why did I pull the fake data for the response from a separate helper method in the request like `$request->getRequest()->getFakeData()`? Because in almost every fake data I want to use some dynamic data from within the Request object, e.g. `$this->id` (lots of APIs return part of the submitted data in their responses). It should not just be dumb and return static or random/fake data. If we would skip the abstraction of an extra `getFakeData()` method, we would end up in even longer callbacks, couldn't use an arrow function, but instead would need to do something like this:

```php
            $this->withMockClient(new MockClient([
                FetchNote::class => function (PendingRequest $pendingRequest) {
                    /** @var FetchNote $request */
                    $request = $pendingRequest->getRequest();
                    return MockResponse::make($request->getFakeData());
                },
                // ...
            ]));
```

The `@var` typehint would be needed, as the IDE does not know about the effective Request object (`$pendingRequest->getRequest()`). So we have now 5 lines per mocked response but would still depend on the `getFakeData()` helper method from within the Request, as we want to access also protected properties or other non-public methods from within the Request.

All-in-all, it just gets completely messy and unmaintainable. For my use cases, the connector's constructor could easily grow to 10+ screen heights, mainly caused by extra bulk like in the example above. But with this PR, we can build all fake data directly in the Request object in a standardized way, and still have the luxury of returning anything from `array|string|MockResponse|Fixture` – fully matching your implementation. It's then just a single line with a single class name to enable the whole mocking per Request, or even just a single line per resource, if we want to group the mocked Requests by resource. Completely readable and maintainable. And I always see the fake data right on the spot, where it belongs to, inside the Request object.

Hope this makes it clear why this is totally needed! I mean, it is super easy to extend it that way in your own project with a simple `HasFakeData` interface for the Requests and a `HasMockHelper` trait for the connector. But we wanted to share that work and it's a non breaking change / new feature.

Sammyjo20:
Hey @onlime 

Thank you so much for the detailed explanation of your use case, that helps a lot. I think it's a cool implementation but in the interest of honesty - I wouldn't say I like the idea of mixing test data and architecture (Requests). When I'm looking at a given request class in my application, I shouldn't have to worry at all about the tests. In my opinion, the fixture/mock data should live within the test or a separate test file that is loaded to populate the mock responses. 

I understand the point of standardisation (because well that's Saloon's whole point) however I did try this idea of having mock data inside of requests [very very early on](https://github.com/saloonphp/saloon/blob/v0.0.1/src/Http/SaloonRequest.php#L89) in v1 and I didn't like the idea of mixing. 

I agree with you though that for large test scenarios it can get pretty messy! You could create invokable classes that you use in your tests. For example, I could have a `FetchNoteData` class in my tests directory.

```php
class FetchNoteData
{
     public static function success(): array
     {
           return [...];
     }

     public static function failure(): array
     {
           return [...];
     }
}
```

and then in your tests you would write something like...

```php
Saloon::fake([
     FetchNote::class => FetchNoteData::success(),
]);
```

Also I appreciate the effort of putting this PR together.

Sammyjo20:
I'm going to close this PR for now, thank you for the suggestion.

onlime:
Hey @Sammyjo20

Sorry about the silence and thanks for your response. I would like to pick up the conversation on this topic, as I think you quite misunderstood my use case last time.

It's not (just) about testing, it's about using 100% of my requests of all Saloon integrations in local dev environment or staging, so I always get "working" response data from all requests fired.
I am firing a lot of requests to various API integrations, often directly from some Eloquent model hooks (`updated`, `created`, `saved`, etc model events). So even in development, I want to inspect the log (using my [Laravel HTTP Client Global Logger](https://github.com/onlime/laravel-http-client-global-logger) for detailed logging incl. all request and response headers/body). I couldn't think of anything better than request mocking. And providing fake date from the Request object itself makes it so much easier, as I can access all local properties directly, without doing the work twice (previously, when using plain Laravel Http Client, I had to parse the path, and inspect the request, sometimes even parsing the body of the request,... all bulk without any advantages, just a complete mess.)

I now have a larger project in development that integrates 6 different APIs as Saloon integrations. All of them use this `HasFakeData` interface and all connectors grab the request mocks from the resources, which group all request mocks. And I depend so much on local properties to build smart fake data, so doing that directly in the requests feels definitely right.

It's quite a wild code base where I have used Saloon extensively. If you are ready to look into it, I am glad to share it with you in a private repo (just the Saloon integration's subtree, as otherwise it's an internal project for my own hosting company). Would you find some time to glance over it and comment? I am sure you will understand my use case then. And that nice implementation which @pascalbaljet built for me in this PR then hopefully may also be of interest to you or other Saloon users.

Cheers,
Pipo

PR #332, opened 2023-11-21
Fix | V3 - Remove array filter on multipart builder

### LINKED ISSUE #331: InvalidArgumentException: A 'contents' key is required
When attempting to send multiple values using Multipart Form Body, I encounter an error. While this functionality worked seamlessly in version 2, it appears to be malfunctioning in version 3.

Issue: https://flareapp.io/share/qm1e8Dwm

Sammyjo20:
Fixes #331 

Sometimes APIs will accept an empty string or `null` for the contents of Multipart Values. Previously if you had tried to provide an empty string, you would get an error from Guzzle. This was because of an `array_filter()` that I had added previously. 

I have now also written a real-world test that makes an API call to our test service to ensure real uploads work.


Sammyjo20 (commented):
Self review

Issue #254, opened 2023-07-10
Allow sending duplicated keys in multipart body


bohemima:
I'm almost done with implementing an API that allows for sending duplicated keys (specifically in a multipart in my case).

Guzzle allows for this using this syntax:

```
    'multipart' => [
        [
            'name'     => 'tag',
            'contents' => 'Tag1'
        ],
        [
            'name'     => 'tag',
            'contents' => 'Tag2'
        ],
        [
            'name'     => 'flattery',
            'contents' => 'Cool library still!'
        ]
    ]
```

Not sure how much of a re-write this is in the library it self, but it seems like keys are used extensively from looking at tests, is this something you have thought about before perhaps?

I guess this could be resolved in multiple ways, allowing value to be an array maybe? `MultipartValue(name: 'tag', value: ['Tag1', 'Tag2'])`, as MultipartBodyRespository seems to rely a lot on keys, perhaps this is a better way than rewriting ?

Sammyjo20:
Hey @bohemima many thanks for posting this and bringing this to my attention. There are always APIs that surprise me! I had assumed (wrongly) that a given name should only appear once. 

Unfortunately, what Saloon does, is it will accept a key-less array, but it will add the keys to the array. The only reason I did this was to support retrieving the data back with `$request->body()->get('tag')`. In my opinion, I don't think we need to key the arrays internally. I can't tell if this is going to be a breaking change or not. I could change the key searching to a better, array-search feature? There's a small chance that people are doing `$request->body()->all()` which will be different, so I think it will be a breaking change.

I'm more than  happy to change this in v3 and onwards, and I'll put it in my list of things to change. In the meantime, a workaround could be to use the config and pass this Guzzle configuration in directly into your request instead of using Saloon's body methods.

In your request class:

```php
protected function defaultConfig(): array
{
    return [
         'multipart' => [
        [
            'name'     => 'tag',
            'contents' => 'Tag1'
        ],
        [
            'name'     => 'tag',
            'contents' => 'Tag2'
        ],
        [
            'name'     => 'flattery',
            'contents' => 'Cool library still!'
        ]
    ];
}
```

Really sorry about this!

bohemima:
No worries, I will give your workaround a try tomorrow, many thanks for pointing it out.

I guess the same issue exist when using defaultQuery since it's an keyed array?

I really like the approach you've taken with Saloon, especially the testability. Keep up the good work!

Sammyjo20:
[...]

PR #340, opened 2023-11-30
Fix | V3 - Allow multiple multipart values with the same name

### LINKED ISSUE #338: HasMultipartBody and multiple files with the same key
This pure Laravel code is working properly:

```php
use Illuminate\Support\Facades\Http;


$files = [
            [
                'name' => 'files',
                'contents' => fopen($voice->getMedia('samples')->first()->getPath(), 'r'),
                'filename' => 'sample1.mp3',
                'headers' => [
                    'Content-Type' => 'audio/mpeg'
                ]
            ],
            [
                'name' => 'files',
                'contents' => fopen($voice->getMedia('samples')->last()->getPath(), 'r'),
                'filename' => 'sample2.mp3',
                'headers' => [
                    'Content-Type' => 'audio/mpeg'
                ]
            ]
        ];

        $headers = [
            "Accept" => "application/json",
            "xi-api-key" => config('services.eleven_labs.api_key'),
        ];

        $response = Http::withHeaders($headers)
            ->attach($files)
            ->post('https://api.elevenlabs.io/v1/voices/add', [
                'name' => $voice->name,
                'description' => $voice->description,
            ]);
```

But this Saloon code don't work, I get uploaded only last file, not all files with key "files"

```php
class AddVoiceRequest extends Request implements HasBody
{
    use HasMultipartBody;

    /**
     * Define the HTTP method
     *
     * @var Method
     */
    protected Method $method = Method::POST;

    public function __construct(
        protected string  $name,
        protected MediaCollection $samples,
        protected ?string $description,
    )
    {
    }

    protected function defaultBody(): array
    {
        $files = $this->samples->map(function (Media $file) {
            return new MultipartValue('files', fopen($file->getPath(), 'r'));
        })->toArray();

        return [
            new MultipartValue('name', $this->name),
            new MultipartValue('description', $this->description ?? ''),
            ...$files
        ];
    }

    /**
     * Define the endpoint for the request
     *
     * @return string
     */
    {
        return '/v1/voices/add';
    }
}
```

Sammyjo20:
This PR fixes #338 #254

I'm aware that this PR **technically** introduces a breaking change in the way multipart values are keyed behind the scenes. I'm going to proceed with this PR as this fixes an issue that multiple people have been experiencing and the breaking change is unlikely to affect the end user. I take breaking changes very seriously but decided to go with this one.

### DIFF
--- a/src/Repositories/Body/ArrayBodyRepository.php
+++ b/src/Repositories/Body/ArrayBodyRepository.php
@@ -6,6 +6,7 @@ namespace Saloon\Repositories\Body;
 
 use LogicException;
 use InvalidArgumentException;
+use Saloon\Helpers\ArrayHelpers;
 use Saloon\Traits\Conditionable;
 use Psr\Http\Message\StreamInterface;
 use Saloon\Contracts\Body\MergeableBody;
@@ -102,7 +103,7 @@ class ArrayBodyRepository implements BodyRepository, MergeableBody
             return $this->all();
         }
 
-        return $this->all()[$key] ?? $default;
+        return ArrayHelpers::get($this->all(), $key, $default);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Sammyjo20 (commented):
Self review

PR #341, opened 2023-12-02
Feature | Added certificate authenticator

Sammyjo20:
- Also deprecates the old `xml` method on the response in favour of using XML Wrangler
- Also fixes an issue with the stream not being rewound after accessing the body


Sammyjo20 (commented):
Self review

PR #342, opened 2023-12-03
Fix exponential backoff attribute

boryn:
`$useExponentialBackoff` needs to be defined as `null`, as otherwise when `$useExponentialBackoff` is only specified at the connector level, this line:

`$useExponentialBackoff = $request->useExponentialBackoff ?? $this->useExponentialBackoff ?? false;`

will evaluate `false`, as it has default `false` value for the request class (inherited from `HasTries` trait)

### DIFF
--- a/src/Traits/RequestProperties/HasTries.php
+++ b/src/Traits/RequestProperties/HasTries.php
@@ -31,7 +31,7 @@ trait HasTries
      *
      * When true, Saloon will double the retry interval after each attempt.
      */
-    public bool $useExponentialBackoff = false;
+    public ?bool $useExponentialBackoff = null;
 
     /**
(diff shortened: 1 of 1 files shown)


Sammyjo20 (approved):

Sammyjo20:
Thanks @boryn!

boryn:
PS. It's a breaking change between 3.3.0 and 3.3.1. If somebody already defined `public bool $useExponentialBackoff` in connector / request it will demand a slight change in the definition to (nullable `?bool`):
`public ?bool $useExponentialBackoff`

Sammyjo20:
Oh, whoops. Really sorry I didn’t notice that @boryn. I’ll add it to the release notes but it could be considered a hotfix for the recent release.

PR #347, opened 2023-12-07
Feature | Throw an exception when missing HTTP method

Sammyjo20:
This PR improves the default error message if you forget to add a HTTP method to your request.

It will now say: `Your request is missing a HTTP method. You must add a method property like [protected Method $method = Method::GET]`


Sammyjo20 (commented):
Self review

PR #349, opened 2023-12-08
Feature | V3 Support Multiple Authenticators

Sammyjo20:
With previous versions of Saloon, you could only have one authenticator per connector/request. This meant that if an API utilized multiple ways of authentication like a certificate and an `Authorization` header you would have to write it manually. You can now use the `MultiAuthenticator` to use multiple authenticators at once.

```php
protected function defaultAuth(): Authenticator
{
    return new MultiAuthenticator(
         new TokenAuthenticator('example'),
         new HeaderAuthenticator('api-key', 'X-API-Key'),
    );
}
```

This is a backwards-compatible change and it's entirely opt-in.


Sammyjo20 (commented):
Self review

Sammyjo20:
Update - some feedback is suggesting just introduce the new `MultiAuthenticator` authenticator which can be used when multiple authenticators want to be used at the same time.

src/Traits/Auth/AuthenticatesRequests.php
```diff
@@ -32,17 +35,26 @@ protected function defaultAuth(): ?Authenticator
...
-        return $this->authenticator ?? $this->defaultAuth();
+        if (isset($this->authenticator)) {
+            return $this->authenticator;
+        }
+
+        if ($defaultAuth = $this->defaultAuth()) {
+            $this->authenticate($defaultAuth);
+        }
+
+        return $this->authenticator;
```
JonPurvis: Could you perhaps shorten this code to:

```php
return $this->authenticator ?? $this->authenticate($this->defaultAuth()) ?? $this->authenticator;
```

JonPurvis (approved):
Hey @Sammyjo20 

I really like this idea and can see it coming in very useful for Saloon users. I can't see anything in the code that stands out as problematic. My only comment is a suggestion that equals in a bit less code, but absolutely does not need doing if you don't want to.

src/Traits/Auth/AuthenticatesRequests.php
boryn: And isn't `$this->authenticator` duplicated? I can see it the first and the last?

Sammyjo20 (commented):
Self review

src/Traits/Auth/AuthenticatesRequests.php
Sammyjo20: I've decided to revert to the old code and just accept a `MultiAuthenticator` instance instead of an array or an authenticator - keeps it simpler!

PR #350, opened 2023-12-10
Fix | Remove `xml` method deprecation

Sammyjo20:
The `xml` method was still used by some people so I've decided to remove the deprecation but instead write some better descriptions for the methods to help people decide.


Sammyjo20 (commented):
Self review

PR #357, opened 2024-01-03
Remove Redundant Trait

Cbrad24:
Removes the `HasDelay` trait definition on the abstract class `Connector`, which is already provided by the `HasRequestProperties` trait that is in use on this class, making the definition redundant, so it can be safely removed.

As below:
```php
trait HasRequestProperties
{
    use HasHeaders;
    use HasQuery;
    use HasConfig;
    use HasMiddleware;
    use HasDelay;
}
```


Sammyjo20 (approved):
Ah good spot, thank you @Cbrad24 !

PR #359, opened 2024-01-10
Feature | Global Mock Client

Sammyjo20:
This PR introduces a global mock client for all applications, not just for Laravel. This allows you to register a mock client in your test and then execute application code like making a call to a route which might contain a Saloon request. This means you don't have to pass around a `MockClient` like you had to do before.

This PR is part of an exercise I am doing to improve the testing experience in Saloon. I feel that for people not using Laravel, there wasn't a good way to test globally.

This make use of PHP static properties which are the same throughout the PHP process lifecycle, essentially global variables. 

```php
public function test_my_application_can_make_api_call()
{
    $mockClient = MockClient::global([
        SomeRequest::class => MockResponse::make(['example' => 'data']);
    ]);

    $this->post('/some-internal-route/');

    $mockClient->assertSent(SomeRequest::class);
}
```

> **Note**
> You will need to configure `MockClient::destroyGlobal()` between tests either in a global beforeEach or in the setup method.

**Pest**
```php
uses()
    ->beforeEach(fn () => MockClient::destroyGlobal())
    ->in(__DIR__);
```
**PHP Unit**
```php
public function setUp()
{
    MockClient::destroyGlobal();
}
```
### Upsides
- Offers people with a way to test globally without having to pass around a `$mockClient` variable.
- Uses the existing `MockClient` class under the hood so all expectations are the same

### Downsides
- Laravel testing might become slightly more confusing because there's another way to test. I will need to separate the documentation to be able to support all the different types of mock clients.
- Can result in leaky tests if people forget to reset between tests


Sammyjo20 (commented):
Self review

Sammyjo20 (commented):
Self review

PR #360, opened 2024-01-12
Fix | Update PHPDoc with typo 

chengkangzai:
Hi Team! 

Great love with Saloon!

Just spot a typo when i try to fix catch a exception and try to fix it here


Sammyjo20 (approved):
Thank you for fixing this!

PR #362, opened 2024-01-14
Fix | Remove Unnecesary @ throws tags in docblocks

Sammyjo20:
This PR cleans up the usages of `@throws` tags inside of PHP docblocks. I have removed the ones which I didn't think were necessary for the end user to try/catch.


Sammyjo20 (commented):
Self review

PR #363, opened 2024-01-18
Fix | Allow createOAuthAuthenticator method to have a nullable refresh token

### LINKED ISSUE #352: AuthConnector::createOAuthAuthenticator(): Argument #2 ($refreshToken) must be of type string, null given
i dont get a refresh token ( shopify api ) and dont have a fallback. 

should't the $refreshToken be Nullable

```
protected function createOAuthAuthenticator(string $accessToken, ?string $refreshToken, ?DateTimeImmutable $expiresAt = null): OAuthAuthenticator
    {
        return new AccessTokenAuthenticator($accessToken, $refreshToken, $expiresAt);
    }
```

in AccessTokenAuthenticator the refresh is also nullable
```
 public function __construct(
        public readonly string             $accessToken,
        public readonly ?string            $refreshToken = null,
        public readonly ?DateTimeImmutable $expiresAt = null,
    ) {
        //
    }
```

Sammyjo20:
Fixes #352

### DIFF
--- a/src/Traits/OAuth2/AuthorizationCodeGrant.php
+++ b/src/Traits/OAuth2/AuthorizationCodeGrant.php
@@ -161,7 +161,7 @@ trait AuthorizationCodeGrant
     /**
      * Create the authenticator.
      */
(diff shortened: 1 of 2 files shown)


Sammyjo20 (commented):
Self review

PR #364, opened 2024-01-21
Feature | V3 Added a request option to assertSentCount

Sammyjo20:
This PR also deprecates the `assertSentJson` method because it was poorly implemented.


Sammyjo20 (commented):
Self review

PR #368, opened 2024-02-09
Feature | Deprecate Auth Methods

Sammyjo20:
This PR deprecates the following methods on the connector/request
- `withTokenAuth`
- `withBasicAuth`
- `withDigestAuth`
- `withQueryAuth`
- `withHeaderAuth`
- `withCertificateAuth`

I would like to deprecate these methods as they were originally introduced into v1 as a convenient helper method to quickly add auth to an existing connector or request, but I feel that it goes against Saloon’s re-usability pattern. 

In my opinion, authentication should always be defined in the defaultAuth method of the connector or request however this auth can still be applied using the authenticate method.

### DIFF
--- a/src/Traits/Auth/AuthenticatesRequests.php
+++ b/src/Traits/Auth/AuthenticatesRequests.php
@@ -83,7 +83,7 @@ trait AuthenticatesRequests
     /**
      * Authenticate the request with a query parameter token.
      *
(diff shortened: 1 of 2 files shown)


Sammyjo20 (commented):
Self review

Sammyjo20 (commented):
Self review

Issue #369, opened 2024-02-11
Dots in parameters get converted to underscores because of parse_str


jhard:
I've banged my head against the wall over this for the past hour until I looked at curl's debug output and really paid attention and noticed that `https://api.cloudflare.com/client/v4/zones?account.id=...` becomes `https://api.cloudflare.com/client/v4/zones?account_id=...`, so Cloudflare is correct in ignoring my filtering and I'm lucky I haven't provided some very, very constructive feedback on their API documentation yet.

In the trait `ManagesPsrRequests` there's
```
        parse_str($uri->getQuery(), $existingQuery);
        return $uri->withQuery(
            http_build_query(array_merge($existingQuery, $this->query()->all()))
        );
```
and the docs confirm:
>  Because variables in PHP can't have dots and spaces in their names, those are converted to underscores. Same applies to naming of respective key names in case of using this function with result parameter. 

I've found that I can do 
```
    public function query() : ArrayStoreContract
    {
        $query = new ArrayStore();
        $query->set([ "account.id" => $this->accountId ]);

        return $query;
    }
```
in the Request as a workaround, but that feels a lot more clunky than just having
```
    public function resolveEndpoint() : string
    {
        return "/zones?account.id={$this->accountId}";
    }
```

Guzzle does some explode-based parsing (`GuzzleHttp\Psr7\Uri::getFilteredQueryString`) which looks a little sketchy, but it should work just as well. `parse_str` doesn't support `;` as a delimiter either, so I assume it should be fine to switch to something like that unless there's a conscious decision behind using `parse_str` that I'm not aware of.

Sammyjo20:
Hey @jhard 

Thank you for making me aware of this issue and I'm sorry for your frustration that this caused you! Does this still happen when you use the `defaultQuery` method on the connector/request?

This is the recommended way to add query parameters:

```php
class YourRequest extends Request
{
    // ...

    protected function defaultQuery(): array
    {
        return [
             'account.id' => $this->accountId,
        ];
    }
}
```

However, I do support putting query parameters in the URL - so I'm going to look at a much better query string parser! Should have a PR up soon.

Sammyjo20:
Do you mind having a quick look at this PR please @jhard? https://github.com/saloonphp/saloon/pull/370 Just want someone to double-check the new implementation. I think it's a bit better than the explode-based passing as if someone wrote a query parameter like `foo==bar` it will still be parsed as `foo=bar`.

jhard:
[...]

PR #370, opened 2024-02-12
Fix | Better Query Parameter Parsing

Sammyjo20:
This PR fixes issue #369 

The old query parameter parser was just using `parse_str` however this was converting dots into underscores. This PR introduces a much better tokenization-parsing algorithm which does not convert characters like dots and semi-colons into underscores.


Sammyjo20 (commented):
Self review

Sammyjo20 (commented):
Self review

PR #371, opened 2024-02-12
Fix | Allow Return False In hasRequestFailed

Sammyjo20:
Closes #314 

This PR inherits a lot of the changes that @gdebrauwer originally proposed, so thank you very much for figuring out that issue!

This PR changes the way the `hasRequestFailed` method works. Previously it would only work if you returned `true`, useful for APIs that have an invalid HTTP status code. However, @gdebrauwer discovered that if you wanted to tell Saloon that a request has not failed even if it has a valid HTTP status, this did not work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 .php-cs-fixer.cache
 .phpunit.result.cache
 tests/Fixtures/Saloon/Testing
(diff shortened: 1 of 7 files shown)


Sammyjo20 (commented):
Self review

PR #377, opened 2024-02-23
Feature | Added dataUrl for Response

guanguans:


Sammyjo20:
Hey, @guanguans thank you for submitting this PR - I think this is a cool addition to Saloon! I'm going to have a quick look over the PR but I think the implementation is solid.

src/Http/Response.php
```diff
@@ -333,6 +333,14 @@ public function dom(): Crawler
         return new Crawler($this->body());
     }
 
+    /**
+     * Convert the response to a data URL
+     */
+    public function dataUrl(): string
+    {
+        return 'data:'.$this->psrResponse->getHeaderLine('content-type').';base64,'.base64_encode($this->body());
```
Sammyjo20:
```suggestion
        return 'data:'.$this->psrResponse->getHeaderLine('Content-Type').';base64,'.base64_encode($this->body());
```

Sammyjo20 (changes requested):
Just one suggestion - would it be better to used the camel-case variant for `Content-Type`? I guess it's up to the PSR-7 implementation behind the scenes on how it handles it, but I'm pretty sure Guzzle normalises it anyway.

src/Http/Response.php
guanguans: 👍 It has been fixed.

Sammyjo20 (approved):
Looks good! I will release this later. Thanks again!

PR #379, opened 2024-02-25
Feature | Deprecate Send And Retry Method

Sammyjo20:
This PR deprecates the old `sendAndRetry` method in Saloon as it goes against Saloon's class-based pattern. You are still able to define retry functionality but you can now do this from within the connector or request class by using the `$tries` property.


Sammyjo20 (commented):
Self review

PR #396, opened 2024-04-07
Proposal: Generic types for DTO casting

michaeldyrynda:
I've added annotations for the `dto` and `dtoOrFail` methods, enabling passing of a `class-string` to each method, in order to get proper type suppose when working with the response.

The new `$dto` parameter is optional to enable backwards-compatibility, but means that you can call the method with a FQCN passed as an argument, in order to get type completion.

Omitting the argument continues as currently.

I was inspired by @ash-jc-allen's [blog post](https://ashallendesign.co.uk/blog/data-transfer-objects-dtos-in-php), which passed the `FQCN` to this method, which initially looked unnecessary to me.

```diff
public function getRepo(string $owner, string $repo): Repo
{
    return $this->client()
        ->send(new GetRepo($owner, $repo))
+       ->dtoOrFail(Repo::class);
}
```

This implementation conforms to PHPStan requirements, so I believe it to be _technically_ correct, but I wasn't actually able to verify it in my editor (Neovim).

For it to function (for me), I had to tweak the annotation slightly:

```diff
  * @template T of object
+ * @param class-string<T>|null $dto
- * @return ($dto of class-string<T> ? T : object)
+ * @return T
```

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -287,8 +287,11 @@ class Response
 
     /**
      * Cast the response to a DTO.
+     *
+     * @template T of object
+     * @return ($dto is class-string<T> ? T : object)
      */
-    public function dto(): mixed
+    public function dto(?string $dto = null): mixed
     {
         $request = $this->pendingRequest->getRequest();
         $connector = $this->pendingRequest->getConnector();
@@ -304,8 +307,11 @@ class Response
 
     /**
      * Convert the response into a DTO or throw a LogicException if the response failed
+     *
+     * @template T of object
+     * @return ($dto is class-string<T> ? T : object)
      */
-    public function dtoOrFail(): mixed
+    public function dtoOrFail(?string $dto = null): mixed
     {
         if ($this->failed()) {
             throw new LogicException('Unable to create data transfer object as the response has failed.', 0, $this->toException());
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -76,7 +76,7 @@ class Response
     /**
      * Create a new response instance.
      */
-    public function __construct(ResponseInterface $psrResponse, PendingRequest $pendingRequest, RequestInterface $psrRequest, Throwable $senderException = null)
+    public function __construct(ResponseInterface $psrResponse, PendingRequest $pendingRequest, RequestInterface $psrRequest, ?Throwable $senderException = null)
     {
         $this->psrRequest = $psrRequest;
         $this->psrResponse = $psrResponse;
@@ -193,7 +193,7 @@ class Response
     /**
      * Get the JSON decoded body of the response as an array or scalar value.
      *
-     * @param array-key|null $key
+     * @param  array-key|null  $key
      * @return ($key is null ? array<array-key, mixed> : mixed)
      */
     public function json(string|int|null $key = null, mixed $default = null): mixed
@@ -214,7 +214,7 @@ class Response
      *
      * Alias of json()
      *
-     * @param array-key|null $key
+     * @param  array-key|null  $key
      * @return ($key is null ? array<array-key, mixed> : mixed)
      */
     public function array(int|string|null $key = null, mixed $default = null): mixed
@@ -265,9 +265,10 @@ class Response
      * Get the JSON decoded body of the response as a collection.
      *
      * Requires Laravel Collections (composer require illuminate/collections)
+     *
      * @see https://github.com/illuminate/collections
      *
-     * @param array-key|null $key
+     * @param  array-key|null  $key
      * @return \Illuminate\Support\Collection<array-key, mixed>
      */
     public function collect(string|int|null $key = null): Collection
@@ -289,6 +290,7 @@ class Response
      * Cast the response to a DTO.
      *
      * @template T of object
+     *
      * @return ($dto is class-string<T> ? T : object)
      */
     public function dto(?string $dto = null): mixed
@@ -309,6 +311,7 @@ class Response
      * Convert the response into a DTO or throw a LogicException if the response failed
      *
      * @template T of object
+     *
      * @return ($dto is class-string<T> ? T : object)
      */
     public function dtoOrFail(?string $dto = null): mixed
@@ -400,7 +403,7 @@ class Response
     /**
      * Execute the given callback if there was a server or client error.
      *
-     * @param callable($this): (void) $callback
+     * @param  callable($this): (void)  $callback
      * @return $this
      */
     public function onError(callable $callback): static
@@ -460,6 +463,7 @@ class Response
      * Throw an exception if a server or client error occurred.
      *
      * @return $this
+     *
      * @throws \Throwable
      */
     public function throw(): static
@@ -504,7 +508,7 @@ class Response
     /**
      * Save the body to a file
      *
-     * @param string|resource $resourceOrPath
+     * @param  string|resource  $resourceOrPath
      */
     public function saveBodyToFile(mixed $resourceOrPath, bool $closeResource = true): void
     {
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -320,7 +320,7 @@ class Response
             throw new LogicException('Unable to create data transfer object as the response has failed.', 0, $this->toException());
         }
 
-        return $this->dto();
+        return $this->dto($dto);
     }
 
     /**
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -291,14 +291,20 @@ class Response
      *
      * @template T of object
      *
-     * @return ($dto is class-string<T> ? T : object)
+     * @return ($type is class-string<T> ? T : object)
      */
-    public function dto(?string $dto = null): mixed
+    public function dto(?string $type = null): mixed
     {
         $request = $this->pendingRequest->getRequest();
         $connector = $this->pendingRequest->getConnector();
 
         $dataObject = $request->createDtoFromResponse($this) ?? $connector->createDtoFromResponse($this);
+
+        if (! is_null($type) && ! is_null($dataObject) && $dataObject::class !== $type) {
+            throw new InvalidArgumentException(
+                message: sprintf('The class-string provided [%s] must match the class-string returned by the connector/request [%s].', $type, $dataObject::class),
+            );
+        }
 
         if ($dataObject instanceof WithResponse) {
             $dataObject->setResponse($this);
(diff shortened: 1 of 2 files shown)


juse-less:
@michaeldyrynda Have you tried with both Intelephense and Phpactor?

My experience is that Phpactor works far better.

It _shouldn't_ be the reason, but I also realise that the return type of the DTO methods is mixed rather than object.
Have you tried change it to object, as well as your conditional fallback return type to mixed?

In terms of changing this type, static analysis tools will start failing for those of us who use custom responses and override the DTO methods directly.
I'm not sure if that strictly qualifies as a breaking chance, though.

I have actually managed to solve DTOs, by defining them in the request, and have them propagate to the DTO methods in the response.
But PhpStorm (that I primarily tested in) doesn't work with cross-file generics like that.
So I gave up rather than propose it.
PHPStan was satisfied, though, and I could perfectly fine use `\PHPStan\dumpType()` the correct type from the DTO methods.

I can commit and push it if someone is interested anyway.

ash-jc-allen:
Hey all! I was just looking through my blog post and agree that it was unecessary for me to add that parameter. I think it was a typo on my end that slipped through my proof reading haha 🤦

Although, I do like the idea of getting some kind of autocomplete and static analysis in my IDE, so I'd be a fan of this addition 🙂

michaeldyrynda:
> I have actually managed to solve DTOs, by defining them in the request, and have them propagate to the DTO methods in the response.

Agree; I typically use resource-based SDK builds like the docs suggest, so do my typed returns on the connector, rather than on the request itself.

I've tried with phpactor and it makes no difference sadly (though I'll be making the switch to that permanently regardless, it provides better diagnostics than intelephense!), which is a shame.

Wish there was a better way of handling this, as having to annotate things by hand is grim.

src/Http/Response.php
```diff
@@ -287,8 +287,11 @@ public function collect(string|int|null $key = null): Collection
 
     /**
      * Cast the response to a DTO.
+     *
+     * @template T of object
+     * @return ($dto is class-string<T> ? T : object)
```
juse-less:
```suggestion
     *
     * @template T of object
     * @param class-string<T>|null $dto
     * @return ($dto is null ? object : T)
```

src/Http/Response.php
```diff
@@ -304,8 +307,11 @@ public function dto(): mixed
 
     /**
      * Convert the response into a DTO or throw a LogicException if the response failed
+     *
+     * @template T of object
+     * @return ($dto is class-string<T> ? T : object)
```
juse-less:
```suggestion
     *
     * @template T of object
     * @param class-string<T>|null $dto
     * @return ($dto is null ? object : T)
```

juse-less (commented):
@michaeldyrynda It doesn't make sense to me why it wouldn't work, because it's a fairly simple approach, and not cross-file.

So I wonder if it's the fact that you never defined the param type?
I can't test right now, but see if something like I propose does.

Edit: re-reading your PR message, the return line you remove (when it doesn't work for you) is using _of_, rather than _is_.
It does have the param annotation, that I suggest below, but, the code for the PR didn't.
Could just be typos, though.

juse-less:
I did some tests with the proposed solution, and it works with Phpactor in Neovim, albeit with some adjustments.
I haven't tried Intelephense, though.

I set up a fresh Neovim using NvChad, and then just installed Phpactor.

Interestingly enough, PhpStorm doesn't handle it at all.
But it's also in-line with my experience of the two - Phpactor being far more PHP-intelligent than PhpStorm.

<details>
<summary>Screenshots of relevant code in Neovim and PhpStorm</summary>

**Neovim:**
CleanShot 2024-04-12 at 22 41 46@2x

**PhpStorm:**
CleanShot 2024-04-12 at 22 48 44@2x

</details>

michaeldyrynda:
Well I'll be a monkey's uncle; I tried this again today and it seems to be working as per the PR.

CleanShot 2024-04-13 at 13 02 16

Unfortunately, making the change as @juse-less suggested raises the ire of PHPStan.

CleanShot 2024-04-13 at 13 06 07

So it would seem that the annotation is correct, and functions with both PHPStan and Intelephense. I believe there's a PHPStan plugin for PhpStorm that should enable intelligence for this? 

As it stands, the PR is correct, appeases static analysis, as well as completion for the two intelligence tools. Given that the annotations are PHPStan ones, I think it's reasonable for it to only work in that context.

Whether or not this is considered breaking, however, is another story.

juse-less:
Ohh. Right.
I believe it's complaining when passing null (or nothing), because T doesn't exist without the class string.
So the check should be to check if `$dto` is null, and, if not, then return T.
```php
/** @return ($dto is null ? object : T) */
```

I've always gotten complaints that the template has to be used on the param.
So interesting that it works.
It's entirely possible it's changed more recently and I just haven't checked.

In terms of PHPStan in PhpStorm, it's bundled and didn't make a difference in my screenshot.
But that plugin seems quirky. For things like class-string, you need the bundled Psalm plugin to be active, f.ex.
I do have both active, but no dice.

If it works as proposed, then I don't really see any issues with it per se.
Just that PhpStorm won't have autocompletion.

Sammyjo20:
Hey @michaeldyrynda Thank you so much for this PR, I really like this idea and thank you for contributing too @juse-less ❤️

How do we feel about implementing a type check so if someone accidentally puts a different class in, it throws an error? Or even passing the class as a second argument like: 

```php
createDtoFromResponse(Response $response, string $classString)
```

The latter could open up cool DTO casting like `match` statements based on the class string 🤔 What do you think? I'm happy to keep it simple though

Sammyjo20 (approved):
Awesome PR, thanks @michaeldyrynda!

Sammyjo20:
Hmm this is really strange, the DTO type checking isn't working on my local machine (Running the latest PHPStorm 2024.1.2) was this actually still in discussion?

juse-less:
@Sammyjo20 I wouldn't say it's a discussion.
It's a limitation in PhpStorm itself.

Adding this shouldn't break anything in PhpStorm, from I've seen.
But will add autocompletion for other editors and IDEs.
Once PhpStorm adds the support, it should start working there as well.

I'm not sure there's any tricks that can be done to make it work in PhpStorm.

I can check the things I tested a while back, to see how it works in PhpStorm and Neovim.
But in its current state, it'd have to be merged to the next major.

Sammyjo20:
Thanks for the information @juse-less that's certainly a shame as it's the IDE I love to use. I'm still willing to release this PR as it is as it will help PHPStan analysis and as you mentioned, other editors.

juse-less:
I switch between PhpStorm and Neovim quite a lot, depending on what I do, and where.
I see incompatibilities like these, almost always in PhpStorm.
It's indeed a shame. 😕

The big upside of Michaels proposed solution is that it's so simple.
And it doesn't need much code changes. Neither in Saloon, nor SDKs.

The things I tried for v2, when we added more types and generics, got quite big and cumbersome to use in both Saloon itself, but also SDKs.
The biggest problem being propagating the type/generic across files in a way that all editors and tools understand.

juse-less:
@michaeldyrynda Can you just double-check that the changes proposed in the PR works?

The reason I'm asking is because I tested them just now (your latest commit, without the style changes, from the diff view).
It doesn't work for me in neither Intelephense, nor Phpactor.
<details>
<summary>Changes tested, and screenshot</summary>

```php
/**
 * Cast the response to a DTO.
 * 
 * @template T of object
 *
 * @return ($dto is class-string<T> ? T : object)
 */
public function dto(?string $dto = null): mixed
```

As seen below, it won't suggest `getMyProperty()` method, nor the `myProperty` property.
Not even typing out 'my' works for me.

-------------
Intelephense:
-------------
CleanShot 2024-06-09 at 20 18 40@2x

CleanShot 2024-06-09 at 20 18 32@2x

---------
Phpactor:
---------
CleanShot 2024-06-09 at 20 29 09@2x

CleanShot 2024-06-09 at 20 30 19@2x
</details>

However, when I use my amended proposed changes, both of them works.
As in; I get autocompletion.
<details>
<summary>Changes tested, and screenshot</summary>

```php
/**
 * Convert the response into a DTO or throw a LogicException if the response failed
 * 
 * @template T of object
 *
 * @param class-string<T>|null $dto
 *
 * @return ($dto is null ? object : T)
 */
public function dto(?string $dto = null): mixed
```

As seen below, it suggests the `getMyProperty()` method, and the `myProperty` property.
And only them, even without typing out 'my'.

-------------
Intelephense:
-------------
CleanShot 2024-06-09 at 20 15 52@2x

---------
Phpactor:
---------
CleanShot 2024-06-09 at 20 31 36@2x
</details>

I've tested with a fresh NvChad installation, as well as my own config.
Both in Neovim 0.10.

It still doesn't work for me in PhpStorm, though.

In terms of PHPStan, I don't get any errors related to it.
I applied my changes on the latest v3, and ran PHPStan on it.
<details>
<summary>Screenshot of PHPStan</summary>

CleanShot 2024-06-10 at 00 49 56@2x
</details>

Sammyjo20:
Hey @michaeldyrynda love this PR, any chance you could take a look at this comments and see if you can fix the IDE issues Juse raised?

Gummibeer:
Just an idea - shouldn't it be possible to get the return type of the `createDtoFromResponse` method and assign that to the template to get rid of the manual return type definition and just have it working with all currently existing implementations? 🤔
It's more work as the generic has to be passed around several classes. But as it's available as the return already it should be possible.

juse-less:
@Gummibeer that's how i solved it a while ago.
But I also had to use the response itself.

Technically it works, except for PhpStorm.
But I couldn't get it to work without breaking CIs.

I tried quite a few combinations, dumping types, but couldn't get it to work without template on the response class.

There's also another issue; the connector also use the trait with the `createDtoFromResponse()` method.
The `Response::dto()` method checks both, which made it a bit harder.

It'd be great if someone can figure out a way to make it work.

Sammyjo20:
I wonder if this is worth testing with PHPStan v2 to see if there is any improvement?

Sammyjo20:
I've just opened a PR for a new way of using DTOs in Saloon - would love to get your feedback/generic testing on this if you're interested! https://github.com/saloonphp/saloon/pull/488

Sammyjo20:
I'm going to close this PR as I feel #488 will improve this process. Thanks for all the contributions so far guys! You're all legends.

PR #400, opened 2024-04-11
Feature | Add `assertSentInOrder` assertion to MockClient

DaniTulp:
In Laravel there's quite a neat assertion on the Http client called `assertSentInOrder`, it kind of does what it says on the tin, it allows you to check that the requests that are made are sent in the same order as you expect them.

To me this opens up quite a few testing scenarios like this:
```php
$mockClient->assertSentInOrder([
            function (FirstRequest $request) {
                // some specific assertion on the first request + type info
                return true;
            },
            function (SecondRequest $request) {
                // some specific assertion on the second request + type info
                return true;
            },
        ])
```

### DIFF
--- a/src/Http/Faking/MockClient.php
+++ b/src/Http/Faking/MockClient.php
@@ -263,7 +263,6 @@ class MockClient
      * Assert that given requests were sent in order
      *
      * @param array<\Closure|class-string<Request>|string> $callbacks
-     * @return void
      */
     public function assertSentInOrder(array $callbacks): void
     {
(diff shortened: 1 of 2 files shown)


Sammyjo20:
Thanks for this PR @DaniTulp and sorry for the delay. I will be looking at this properly soon, just wanted to give you a response. I will need a little more time as I'd need to update the documentation.

DaniTulp:
@Sammyjo20 no problem, let me know if you need help with that!

Sammyjo20 (approved):
Awesome PR, thanks @DaniTulp!

PR #403, opened 2024-04-18
Change guzzle client creation (linked to PR in laravel-plugin)

gdebrauwer:
I noticed the `laravel-http-sender` package is being sunset, but I still want requests/responses to be logged in Telescope. Then I discovered this is possible with the following package: https://github.com/huzaifaarain/telescope-guzzle-watcher/tree/main

The problem is that this package only works if the guzzle client is created using the service container. That is currently not the case.

This PR and PR https://github.com/saloonphp/laravel-plugin/pull/62 make that possible with minimal changes.

### DIFF
--- a/src/Http/Senders/GuzzleSender.php
+++ b/src/Http/Senders/GuzzleSender.php
@@ -66,24 +66,37 @@ class GuzzleSender implements Sender
      */
     protected function createGuzzleClient(): GuzzleClient
     {
-        // We'll use HandlerStack::create as it will create a default
-        // handler stack with the default Guzzle middleware like
-        // http_errors, cookies etc.
-
-        $this->handlerStack = HandlerStack::create();
-
         // Now we'll return new Guzzle client with some default request
         // options configured. We'll also define the handler stack we
         // created above. Since it's a property, developers may
         // customise or add middleware to the handler stack.
 
-        return new GuzzleClient([
+        return new GuzzleClient($this->getGuzzleClientConfig());
+    }
+
+    /**
+     * Get Guzzle client config
+     */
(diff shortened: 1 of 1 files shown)


Sammyjo20 (approved):
Looks good, I don't think this would cause any issue with backwards compatibility. Thanks @gdebrauwer !

src/Http/Senders/GuzzleSender.php
```diff
@@ -66,24 +66,37 @@ public function getFactoryCollection(): FactoryCollection
...
-        // We'll use HandlerStack::create as it will create a default
-        // handler stack with the default Guzzle middleware like
-        // http_errors, cookies etc.
-
-        $this->handlerStack = HandlerStack::create();
-
         // Now we'll return new Guzzle client with some default request
         // options configured. We'll also define the handler stack we
         // created above. Since it's a property, developers may
         // customise or add middleware to the handler stack.
```
patrickcarlohickman, line 72: These comments should probably be updated a little bit. "Now" doesn't make sense anymore and the handler stack is not being created above anymore.

pajkho:
What is the status on this? Will it be merged in?

Sammyjo20:
The `saloonphp/laravel-http-sender` has been sunset but it will soon be replaced with a new package that @craigpotter is working on, Barstool - which will record all Saloon requests/responses for you. You can manually change the way the sender is resolved to use the service container with this guide:

https://docs.saloon.dev/conclusion/how-to-guides/improving-speed-with-laravel

pajkho:
@Sammyjo20 I just wanted to check the status on this? What is the best way currently to get requests logged in Telescope?

Issue #418, opened 2024-05-22
Connector object() throws an exception but json returns null|object?


Braunson:
When attempting to get the data from a connector's request, we can use `->json()` which will return null if it fails/nothing is available. However if we use `->object()` it will throw an exception on error.

Shouldn't the `object()` method also follow the same expectation as `json()`? I would expect a null/empty array response instead of an exception.

Secondly any reason we cannot provide a key/default in the object like we do with `json()` and `array()`?

If this is something that the maintainer feels is necessary. I can put in a PR for it.

Sammyjo20:
Fixed in #420 thanks for the PR @Braunson !

PR #420, opened 2024-05-23
Feature | Added support to Response object method to now accept a dot notation like json()

Braunson:
- Added support to Response object method to now accept a dot notation like json()
- Fixed grammar issue
- Added a new `ObjectHelpers` class
- Added test related to retrieving nested array in object via dot notation key
- Should be backward compatible

Related to #418

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -10,6 +10,7 @@ use SimpleXMLElement;
 use Saloon\Traits\Macroable;
 use InvalidArgumentException;
 use Saloon\Helpers\ArrayHelpers;
+use Saloon\Helpers\ObjectHelpers;
 use Saloon\XmlWrangler\XmlReader;
 use Illuminate\Support\Collection;
 use Saloon\Contracts\FakeResponse;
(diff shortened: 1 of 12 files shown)


src/Http/Response.php
```diff
@@ -53,6 +54,13 @@ class Response
      */
     protected array $decodedJson;
 
+    /**
+     * The decoded JSON response object.
+     *
+     * @var object<object-key, mixed>
+     */
+    protected object $decodedJsonObject;
```
patrickcarlohickman: This variable shouldn't be typehinted as an `object`. The json decoded result may not necessarily be an object. For example, if the response is `[]`, this will decode as an array, and throw an error in the `object()` method. This was one of the original issues with the return typehint of the `object()` method being `object`; it would throw an error when the API returned a properly indexed array.

Braunson: Ahh yeah, updated.

Sammyjo20 (approved):

Sammyjo20 (approved):
Great contribution, thanks @Braunson!

PR #422, opened 2024-05-31
Add client credentials grant with basic auth

patrickcarlohickman:
The currently supported OAuth2 client credentials grant includes the client_id and the client_secret in the body of the request. Some APIs require that the client_id and client_secret are sent using the HTTP Basic Authentication scheme. This PR adds a new client credentials request that supports the basic authentication method instead of the request body method.

The request only adds new files and should not cause any backwards compatibility issues.

RFC reference: https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1

Implementation Notes:
- The new `GetClientCredentialsTokenBasicAuthRequest` was copied from the existing `GetClientCredentialsTokenRequest`. The `defaultBody()` method was updated to remove the `client_id`/`client_secret` from the body, and the `defaultAuth()` method was added to implement the basic authentication.
- The new `ClientCredentialsBasicAuthGrant` trait uses the existing `ClientCredentialsGrant` trait, and just redefines the `resolveAccessTokenRequest()` method to use the new request file.
- A new test was added to ensure the request body and Authentication header are as expected.

### DIFF
--- a/src/Http/OAuth2/GetClientCredentialsTokenBasicAuthRequest.php
+++ b/src/Http/OAuth2/GetClientCredentialsTokenBasicAuthRequest.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Saloon\Http\OAuth2;
+
+use Saloon\Enums\Method;
+use Saloon\Http\Request;
+use Saloon\Contracts\Body\HasBody;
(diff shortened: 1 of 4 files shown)


Sammyjo20 (approved):
Awesome PR @patrickcarlohickman thank you! I actually ran into this issue recently with an API, so it's great this will be native now.

Issue #402, opened 2024-04-17
Remove authentication on a specific request


labomatik:
I would like to remove the authentication on a specific request while the connector is using OAuth

The request to the API is the following:
- OAuth request to get the Bearer
- Create a Job -> we are receiving an URL  (Using Bearer)
- PUT a file to the URL  (No Auth should be used)

The last step is actually sending an error when we push to AWS "Only one auth mechanism allowed;" since we are using an API key on that step

Is there a way to remove from the connector the Authentication?

```
        $api = new APIConnector(config('services.api.client_id'), config('services.api.client_secret'));

        $authenticator = $api->getAccessToken([
            "connect/read:jobs",
            "connect/submit:jobs",
            "connect/read:sending-events"
        ]);

        $api->authenticate($authenticator);

        $data = $api->send(new CreateJobSingleFile('01HT2RK56ETAD2GGNRGA4Z03BK', 'test', 'test.pdf'));

        $jobFile = $data->dtoOrFail();

        // We should now post file to $jobFile->fileUploadUrl

        $file = \Storage::disk('local')->path('test.pdf');

        // We should remove the Auth

        $result = $api->send(new AttachFileToJob('01HT2RK56ETAD2GGNRGA4Z03BK', 'test.pdf', $file, $jobFile->fileUploadUrl));

```

jlevers:
Maybe you could do it with a `NullAuthenticator`, something like this:

```php
class NullAuthenticator implements Authenticator
{
    public function set(PendingRequest $pendingRequest): void
    {
    }
}
```

And then when you want to make the auth-free call:

```php
$api->authenticate(new NullAuthenticator());
```

Sammyjo20:
Hey @labomatik I have just released v3.9.0 which includes the above PR made by @patrickcarlohickman. Would you be able to try the `NullAuthenticator` and see if this fixes your issue?

labomatik:
Hello,

Sorry for being late on this... I just validated and it seems that the nullAuthenticator doesn't erase all previous auth mechanism, the API we are using is still pinpointing an issue with a mixed auth...
I added a debug just after the NullAuthenticator and i do see the oauthConfig object part of the connector

![image](https://github.com/saloonphp/saloon/assets/179987/fac13306-0904-4855-b2cd-f539a595b350)

patrickcarlohickman:
@labomatik ,

[...]

Issue #385, opened 2024-03-15
Skipping AuthenticatePendingRequest when fetching a new accessToken using OAuth2 plugins


leroy:
I wanted to fetch the access token using ->getAccessToken() from within the default auth so every request has a valid access token by default.

However, getAccessToken in turn triggers a defaultAuth() call resulting in an infinite loop.

Would it be an idea to skip the AuthenticatePendingRequest somehow when running getAccessToken()?

Willing to propose PR!

```php
public function defaultAuth(): ?TokenAuthenticator
    {
        $cachedToken = cache()->remember('token, now()->addHour(), function() {
            return $this->getAccessToken()->getAccessToken() // Results in recursive loop as underlying request also tries to define defaultAuth();
        });

        return new TokenAuthenticator($cachedToken);
    }
```

Sammyjo20:
Hey @leroy thanks for your patience on my response

This is an interesting one! I'm guessing we probably don't need to call the `defaultAuth()` method for the OAuth2 requests however I could imagine that there might be other ways people need to authenticate so I think we should keep the functionality. 

There's a couple of workarounds I can think of for your situation - first is a custom authenticator class. Within this authenticator class you'll get access to the PendingRequest class so you can ignore specific requests from applying the authentication. For example:

```php
<?php

use Saloon\Http\PendingRequest;
use Saloon\Contracts\Authenticator;

class CustomAuthenticator implements Authenticator
{
    public function set(PendingRequest $pendingRequest): void
    {
        $request = $pendingRequest->getRequest();

        if ($request instanceof GetAccessTokenRequest) {
           return;
        }

        $connector = $pendingRequest->getConnector();

        $cachedToken = cache()->remember('token, now()->addHour(), function() use ($connector) {
            return $connector->getAccessToken();
        });

        $pendingRequest->authenticate(new TokenAuthenticator($cachedToken));
    }
}
```

You can then use this authenticator like you do already:

```php
public function defaultAuth(): CustomAuthenticator
{
    return new CustomAuthenticator;
}
```

Let me know your thoughts on this - the other way I was thinking was pretty much the same but using the `boot()` method on the connector, however I think a custom authenticator is cleaner.

Sammyjo20:
Hey @leroy did the above solution work okay for you?

patrickcarlohickman:
Another option is to use a new connector instance with a null authenticator to get the access token.

```
public function defaultAuth(): ?TokenAuthenticator
{
[...]

PR #423, opened 2024-05-31
Feature | Add a `NullAuthenticator` to overwrite defaultAuth

patrickcarlohickman:
This PR adds a `NullAuthenticator` class that allows a user to use the null object pattern, and provides a way to remove an authenticator that has already been set.

This has been addressed in #402 and [#385](https://github.com/saloonphp/saloon/issues/385#issuecomment-2125073941).

This PR only adds a new file, so it should not have any backwards compatibility issues.

### DIFF
--- a/src/Http/Auth/NullAuthenticator.php
+++ b/src/Http/Auth/NullAuthenticator.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Saloon\Http\Auth;
+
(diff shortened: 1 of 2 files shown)


Sammyjo20 (approved):
Great contribution, thanks @patrickcarlohickman!

jlevers:
Ah, thanks @patrickcarlohickman! I've been meaning to submit a similar PR for a few weeks now. Good stuff :)

PR #424, opened 2024-06-01
Feature | Accept type-hinted requests as `assertSent` parameters

### LINKED ISSUE #417: Asserting a request is sent by closure parameter
Take the following snippet:

```php
MockClient::getGlobal()->assertSent(function (CreateDealRequest $request) {
    expect($request->body()->all())->toMatchArray([
        'title' => '[MGT] Jon Doe / foo - bar / ' . now()->addDay()->format('M d \'y'),
        'status' => DealStatus::OPEN,
    ]);
});
```

This would fail if there was a request other than `CreateDealRequest`. It would be great if `assertSent` could take the request type into account.

Open to PRing that 👍

innocenzi:
Closes https://github.com/saloonphp/saloon/issues/417

This pull request adds support for type-hinting a request class as the first parameter of the closure passed to `MockClient#assertSent`. This allows running this kind of test:

```php
MockClient::getGlobal()->assertSent(function (CreateDealRequest $request) {
    expect($request->body()->all())->toMatchArray([
        'status' => DealStatus::OPEN,
    ]);
});
```

Before this pull request, the test above could only be written using a condition:

```php
MockClient::getGlobal()->assertSent(function (Request $request) {
    if ($request instanceof CreateDealRequest) {
        expect($request->body()->all())->toMatchArray([
            'status' => DealStatus::OPEN,
        ]);

        return true;
    }
});
```

### DIFF
--- a/phpstan.baseline.neon
+++ b/phpstan.baseline.neon
@@ -54,3 +54,18 @@ parameters:
 		 message: "#^Match arm is unreachable because previous comparison is always true.$#"
 		 count: 1
 		 path: src/Http/Pool.php
+
+		-
+		 message: "#^Method Saloon\\\\Http\\\\Faking\\\\MockClient\\:\\:getRequestClass\\(\\) should return class-string|null but returns class-string<Saloon\\\\Http\\\\Request>|Saloon\\\\Http\\\\Request$#"
+		 count: 1
+		 path: src/Http/Faking/MockClient.php
+
+		-
+		 message: "#^Call to an undefined method ReflectionType\\:\\:getName\\(\\)\\.$#"
+		 count: 1
+		 path: src/Http/Faking/MockClient.php
+
+		-
+		 message: "#^Parameter \\#1 \\$function of class ReflectionFunction constructor expects Closure|string, callable\\(\\)\\: mixed given\\.$#"
+		 count: 1
+		 path: src/Http/Faking/MockClient.php
--- a/src/Http/Faking/MockClient.php
+++ b/src/Http/Faking/MockClient.php
@@ -451,14 +451,29 @@ class MockClient
             return false;
         }
 
-        if(! is_null($index)) {
+        if (! is_null($index)) {
             $response = $this->getRecordedResponses()[$index];
             $request = $response->getPendingRequest()->getRequest();
 
             return $closure($request, $response);
         }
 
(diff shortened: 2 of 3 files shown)


innocenzi:
I ignored the phpstan errors as they seem to be false-positives.

juse-less:
For some reason, I can't get GitHub to let me comment on the changes on my tablet, so I'll make a regular comment for now.

The PHPStan errors around the Reflection APIs might need an extra look.
F.ex., the ReflectionFunction only accepts a Closure, but we're passing a callable.
I'm not sure how common it is to pass an invokable object, though.
That should be relatively easy to fix.
```php
new ReflectionFunction($callable(...));
```

I'm not 100 % sure on the ReflectionType for the parameters.
ReflectionType itself doesn't have a `getName()`, nor does it's children except for the ReflectionNamedType.
This may or may not be an issue.
If only one type is defined, it'll work.
If no parameters are defined, then index 0 will also give an error.
But realistically, I feel like these are edge cases.

phpstan.baseline.neon
```diff
@@ -54,3 +54,18 @@ parameters:
...
+
+		-
+		 message: "#^Call to an undefined method ReflectionType\\:\\:getName\\(\\)\\.$#"
+		 count: 1
+		 path: src/Http/Faking/MockClient.php
+
+		-
+		 message: "#^Parameter \\#1 \\$function of class ReflectionFunction constructor expects Closure|string, callable\\(\\)\\: mixed given\\.$#"
+		 count: 1
+		 path: src/Http/Faking/MockClient.php
```
Sammyjo20, line 66:
```suggestion
```

Sammyjo20 (approved):
The PR looks okay to me. I don't think we need these PHPStan exclusions?

phpstan.baseline.neon
innocenzi: Without them, phpstan doesn't pass: 

CleanShot 2024-08-05 at 02 14 13

I believe they are false-positive.

juse-less: They aren't false positives, but the question is if they need addressing or if it should be considered an invalid usage by the user.

Sammyjo20:
Thank you for this PR ✅

innocenzi:
Hey @Sammyjo20, thank you for merging! Have you planned the next Saloon release yet?

PR #425, opened 2024-06-01
Feature | Allow merging data in mocked fixture responses

### LINKED ISSUE #416: Allow partial overriding of fixtures
So I've been using automatic fixture generation to test integration with an API that returns a lot of data.

However, I find myself duplicating some fixtures to edit some information in them so my assertions can pass depending on what I send in the request.

Instead of doing that, it would be great to have the following API:

```php
MockClient::global([
    UpdateDealRequest::class => MockResponse::fixture('pipedrive/deal/update-16685-response', merge: [
        'status' => 'won'
    ]),
]);
```

The `merge` parameter would merge the array into the mocked response, so the following assertion would pass:

```php
expect($response->dto())->status->toBe(Status::WON);
```

I am totally open to PRing that, I created the issue to avoid working on it if it was not going to be accepted 👍

innocenzi:
Closes https://github.com/saloonphp/saloon/issues/416

This pull request allows merging data in fixtures:

```php
MockResponse::fixture('user', merge: [
    'name' => 'Sam Carré',
])
```

This is useful when the mocked response is used by other part of the code that depends on it, and assertions are performed later on the result of that code. 

One (very) simple example of that would be testing if a data transfer object can be created:

```php
$response = connector()->send(new DTORequest, new MockClient([
    MockResponse::fixture('user', merge: [
        'name' => 'Sam Carré',
    ]),
]));

expect($response->dto())
    ->name->toBe('Sam Carré') // Sam Carré instead of Sam
    ->actualName->toBe('Sam')
    ->twitter->toBe('@carre_sam');
```


innocenzi:
One small concern I have is that this code uses `array_merge`, while I'm certain to have use cases where `array_merge_recursive` would be more useful. 

I wanted to stay consistent with implementations of `MergeableBody` which all use `array_merge`. Would you be open to a PR that updates all of these to `array_merge_recursive`?

innocenzi:
@Sammyjo20 do you think you could take a look at that? this would *really* ease testing requests 🙏

innocenzi:
You can amend my previous concern, I refactored the pull request to support dot-notation, which fixes the issue. 

I've been testing this in a work project, `array_merge` was not enough most of the time, and I needed dot-notation support for most fixtures.

Sammyjo20:
Hey @innocenzi thank you for this PR, it looks like a great addition! The only thing I ask is instead of having it as a second argument could we have a method that chains onto the ::fixture() call? Like `MockResponse::fixture()->merge([])` because I've got an upcoming feature that will also share this API so it would be good if it was unified. What do you think?

innocenzi:
That's a nice idea—I updated the PR!

juse-less (approved):

src/Http/Faking/Fixture.php
```diff
@@ -46,7 +63,33 @@ public function getMockResponse(): ?MockResponse
...
+
+            if (! is_array($this->merge)) {
+                return $response;
+            }
+
+            // First, we get the body as an array. If we're dealing with
+            // a `StringBodyRepository`, we have to encode it first.
+            if (! is_array($body = $response->body()->all())) {
+                $body = json_decode($body ?: '[]', associative: true, flags: \JSON_THROW_ON_ERROR);
+            }
```
Sammyjo20, line 91: I think this might be a breaking change for fixtures that are just a string, and not necesarily JSON. I have fixtures in the past which is just plain text. Please could you add a test for checking plain-text fixtures to ensure this isn't a breaking change?

Additionally, please could we allow so people can use recursive arrays **or** dot-notation e.g

```php
MockResponse::fixture(...)->merge([
      'person' => [
            'name' => 'Sam'
      ],
]);
```

Sammyjo20 (changes requested):

src/Http/Faking/Fixture.php
innocenzi: > @Sammyjo20
> I think this might be a breaking change for fixtures that are just a string, and not necesarily JSON.

Gotta admit I'm not sure how to have plain-text fixtures, they automatically save themselves as `.json` and `RecordedResponse#toFile` uses `json_encode`. Could you hint me where to look at to write this test?

That being said, there is a return statement above that condition that checks for the presence of `merge`, so there shouldn't be a breaking change.

&nbsp;


> @Sammyjo20
> Additionally, please could we allow so people can use recursive arrays **or** dot-notation

I've used Laravel's `Arr::set`, which doesn't support that. `data_set` doesn't support it either. We would have to use `array_merge_recursive` or rewrite the function ourselves.

But honestly, that would lead to some unexpected results, eg. appending data to arrays instead of replacing them. This is what happened to me when I first used `array_merge_recursive` when dogfeeding this feature.

A more flexible option would be to allow passing a callback:

```php
MockResponse::fixture(...)->merge(function (array $data) {
    // do whatever with data?
    return array_merge_recursive($data, [
        'person' => [
            'name' => 'Sam'
        ],
    ]);
});
```

innocenzi: I just updated the implementation to add support for closures. If you don't want that, I can revert the commit, no problem 👍

innocenzi:
@Sammyjo20 any chance you could take another look at this?

Sammyjo20:
Hi @innocenzi sorry about the delay. I'm going through all of Saloon's issues and PRs again at the moment and I will update accordingly.

Sammyjo20 (approved):
Thanks for this PR!

PR #429, opened 2024-06-13
Add `Stringable` to OAuthConfig `set` methods.

avenjamin:
This PR adds `|Stringable` to the `set` methods in the OAuthConfig class that accept a string parameter.

In my use case, I'm constructing URLs using the [spatie/url](https://github.com/spatie/url) package and would like to be able to pass the URL directly to the `setAuthorizeEndpoint()` method:

```PHP
$this->oauthConfig()->setAuthorizeEndpoint(
    Url::fromString('https://api.example.com/oauth2/authorize')
    ->withQueryParameters([
        'app_key' => $appKey,
        'response_type' => 'code'
    ])
);
```

The `Url` implements the `Stringable` `__toString()` method so I can either cast to a string or update Saloon to support `Stringable` parameters :smile:

### DIFF
--- a/src/Helpers/OAuth2/OAuthConfig.php
+++ b/src/Helpers/OAuth2/OAuthConfig.php
@@ -8,6 +8,7 @@ use Closure;
 use Saloon\Http\Request;
 use Saloon\Traits\Makeable;
 use Saloon\Exceptions\OAuthConfigValidationException;
+use Stringable;
 
 /**
  * @method static static make()
@@ -73,7 +74,7 @@ class OAuthConfig
      *
      * @return $this
      */
-    public function setClientId(string $clientId): static
+    public function setClientId(string|Stringable $clientId): static
     {
         $this->clientId = $clientId;
 
@@ -93,7 +94,7 @@ class OAuthConfig
      *
      * @return $this
      */
-    public function setClientSecret(string $clientSecret): static
+    public function setClientSecret(string|Stringable $clientSecret): static
     {
         $this->clientSecret = $clientSecret;
 
@@ -113,7 +114,7 @@ class OAuthConfig
      *
      * @return $this
      */
-    public function setRedirectUri(string $redirectUri): static
+    public function setRedirectUri(string|Stringable $redirectUri): static
     {
         $this->redirectUri = $redirectUri;
 
@@ -133,7 +134,7 @@ class OAuthConfig
      *
      * @return $this
      */
-    public function setAuthorizeEndpoint(string $authorizeEndpoint): static
+    public function setAuthorizeEndpoint(string|Stringable $authorizeEndpoint): static
     {
         $this->authorizeEndpoint = $authorizeEndpoint;
 
@@ -153,7 +154,7 @@ class OAuthConfig
      *
      * @return $this
      */
-    public function setTokenEndpoint(string $tokenEndpoint): static
+    public function setTokenEndpoint(string|Stringable $tokenEndpoint): static
     {
         $this->tokenEndpoint = $tokenEndpoint;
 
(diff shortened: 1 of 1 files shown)


src/Helpers/OAuth2/OAuthConfig.php
```diff
@@ -73,7 +74,7 @@ public function getClientId(): string
      *
      * @return $this
      */
-    public function setClientId(string $clientId): static
+    public function setClientId(string|Stringable $clientId): static
     {
         $this->clientId = $clientId;
```
juse-less, line 79:
```suggestion
        $this->clientId = (string) $clientId;
```

src/Helpers/OAuth2/OAuthConfig.php
```diff
@@ -93,7 +94,7 @@ public function getClientSecret(): string
      *
      * @return $this
      */
-    public function setClientSecret(string $clientSecret): static
+    public function setClientSecret(string|Stringable $clientSecret): static
     {
         $this->clientSecret = $clientSecret;
```
juse-less, line 99:
```suggestion
        $this->clientSecret = (string) $clientSecret;
```

src/Helpers/OAuth2/OAuthConfig.php
```diff
@@ -113,7 +114,7 @@ public function getRedirectUri(): string
      *
      * @return $this
      */
-    public function setRedirectUri(string $redirectUri): static
+    public function setRedirectUri(string|Stringable $redirectUri): static
     {
         $this->redirectUri = $redirectUri;
```
juse-less, line 119:
```suggestion
        $this->redirectUri = (string) $redirectUri;
```

src/Helpers/OAuth2/OAuthConfig.php
```diff
@@ -133,7 +134,7 @@ public function getAuthorizeEndpoint(): string
      *
      * @return $this
      */
-    public function setAuthorizeEndpoint(string $authorizeEndpoint): static
+    public function setAuthorizeEndpoint(string|Stringable $authorizeEndpoint): static
     {
         $this->authorizeEndpoint = $authorizeEndpoint;
```
juse-less, line 139:
```suggestion
        $this->authorizeEndpoint = (string) $authorizeEndpoint;
```

src/Helpers/OAuth2/OAuthConfig.php
```diff
@@ -153,7 +154,7 @@ public function getTokenEndpoint(): string
      *
      * @return $this
      */
-    public function setTokenEndpoint(string $tokenEndpoint): static
+    public function setTokenEndpoint(string|Stringable $tokenEndpoint): static
     {
         $this->tokenEndpoint = $tokenEndpoint;
```
juse-less, line 159:
```suggestion
        $this->tokenEndpoint = (string) $tokenEndpoint;
```

src/Helpers/OAuth2/OAuthConfig.php
```diff
@@ -173,7 +174,7 @@ public function getUserEndpoint(): string
      *
      * @return $this
      */
-    public function setUserEndpoint(string $userEndpoint): static
+    public function setUserEndpoint(string|Stringable $userEndpoint): static
     {
         $this->userEndpoint = $userEndpoint;
```
juse-less, line 179:
```suggestion
        $this->userEndpoint = (string) $userEndpoint;
```

juse-less (changes requested):
@avenjamin This could be a breaking change.
I'm not sure if this specific class can/should be seen as a public API.

Anything extending these methods will cause fatal errors as their definitions are now narrower.
I'm not sure how common it is to extend this class or override its methods, though.

If it's considered ok here, you still need to explicitly cast the value to a string, as strict types (which is used in this file) will prevent type juggling.

Sammyjo20:
Thank you for the PR and the suggestions @juse-less. To be honest, I don't think Saloon should support casting strings like this, if I were to implement the PR, I would have to add it everywhere to keep it consistent.

PR #441, opened 2024-07-28
Add `PaymentRequired` Exception

JonPurvis:
This PR adds an Exception for 402 Payment Required so rather than it returning a `ClientException`, it'll now return `PaymentRequiredException`. I'm working with an API that returns this status code and I'd like to be able to handle these in a certain way, this status also seems to be getting more popular these days, so I think it makes sense for it to now have it's own Exception.

Before
Screenshot 2024-07-28 at 23 48 51

After
Screenshot 2024-07-28 at 23 48 04

Sister PR to https://github.com/Sammyjo20/saloon-docs/pull/71

### DIFF
--- a/src/Exceptions/Request/Statuses/PaymentRequiredException.php
+++ b/src/Exceptions/Request/Statuses/PaymentRequiredException.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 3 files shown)


juse-less (approved):

Sammyjo20 (approved):
Awesome PR, thanks Jon! ⭐

PR #454, opened 2024-09-04
Feature | Add methods to dynamically set/unset retry options

viicslen:
This PR adds methods in the `HasTries` trait to allow the retry options to be set dynamically (mainly when using resources)

```php
$this->connector
	->withRetry(tries: 3, intervalMilliseconds: 1000, exponentialBackoff: true, , throwOnMaxTries: true)
	->admin()
	->externalProducts()
	->upsertExternalProduct($productId, $data);


$this->connector
	->withRetry(3)
	->withRetryInterval(1000)
	->withExponentialBackoff()
	->throwOnMaxTries()
	->admin()
	->externalProducts()
	->upsertExternalProduct($productId, $data)
```

This will also allow to unset these options for specific requests

```php
$this->connector
	->withoutRetry()
	->admin()
	->externalProducts()
	->upsertExternalProduct($productId, $data);


$this->connector
	->withRetry(3)
	->withRetryInterval(null)
	->withExponentialBackoff(false)
	->throwOnMaxTries(false)
	->admin()
	->externalProducts()
	->upsertExternalProduct($productId, $data)

$this->connector
	->withRetry(3)
	->withoutRetryInterval()
	->withoutExponentialBackoff()
	->throwOnMaxTries(false)
	->admin()
	->externalProducts()
	->upsertExternalProduct($productId, $data)
```

This is just a draft PR, it would still be missing tests and method comments to follow the other methods in the trait, I will happily add them if you thing this would be a good addition

To Do:
- [ ] Tests
- [ ] Method docblocks

### DIFF
--- a/src/Traits/RequestProperties/HasTries.php
+++ b/src/Traits/RequestProperties/HasTries.php
@@ -60,7 +60,7 @@ trait HasTries
         return $this;
     }
     
-    public function withExponentialBackoff(?bool $exponentialBackoff): static
+    public function withExponentialBackoff(?bool $exponentialBackoff = true): static
     {
         $this->useExponentialBackoff = $exponentialBackoff;
 
@@ -91,6 +91,16 @@ trait HasTries
         }
 
         return $this;
+    }
+
+    public function withoutRetryInterval(): static
+    {
+        return $this->withRetryInterval(null);
+    }
+    
+    public function withoutExponentialBackoff(): static
+    {
+        return $this->withExponentialBackoff(false);
     }
 
     public function withoutRetry(): static
--- a/src/Traits/RequestProperties/HasTries.php
+++ b/src/Traits/RequestProperties/HasTries.php
@@ -52,4 +52,49 @@ trait HasTries
     {
         return true;
     }
+
+    public function withRetryInterval(?int $milliseconds): static
+    {
+        $this->retryInterval = $milliseconds;
+
+        return $this;
+    }
+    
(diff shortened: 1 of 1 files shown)


Sammyjo20:
I've wanted this myself for quite a while, so yes @viicslen I'd love this! Thank you for the suggestion, if you could add tests etc let me know and I'll review it :)

juse-less:
I find the `without*()` methods-, and the `withRetry()` method a bit problematic, for the lack of a better word.

The `withRetry()` method accepts nullable values, but only checks for truey values.
So passing false as argument for the boolean parameters will be ignored.
Same thing if 0 is passed as interval; although this might be undesirable anyway.

The `without*()` methods are a bit inconsistent, in that the integer-based ones sets null.
Whereas the boolean-based ones sets false, even though their values are also nullable.
Without digging too deep into how the retry feature works, [the boolean values are not always considered false when the value is null](https://github.com/saloonphp/saloon/blob/d5bb5d6d72ed33efe19a8463ef259f0aa7e98579/src/Traits/Connector/SendsRequests.php#L39-L40).
Although, I guess it makes sense semantically that non-boolean values are set to null, and boolean values are set to false with `without*()` methods.

But I think it would at least make sense to check if values are not null in `withRetry()`, so that values that are false are still being set.
And maybe `withoutRetry()` should explicitly set the values to false.

It's also possible I'm just overthinking this.

Sammyjo20:
I really appreciate the effort you have gone into this PR, thank you however  I don't feel this is quite right for Saloon. 

I hope you have a great day.

PR #455, opened 2024-09-12
Feature | Added isJson and isXML functions to Response class

mstfblci:
Two new functions have been added to check if the response return is json or xml

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -490,6 +490,26 @@ class Response
     }
 
     /**
+     * Determine if the response is in JSON format.
+     *
+     * @return bool
+     */
+    public function isJson(): bool
+    {
+        return str_contains($this->header('Content-Type'), 'json');
+    }
+
+    /**
+     * Determine if the response is in XML format.
+     *
+     * @return bool
+     */
+    public function isXML(): bool
+    {
+        return str_contains($this->header('Content-Type'), 'xml');
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


SRWieZ:
This could be really useful; in a lot of my Connectors, I do things like this:

```php
public function hasRequestFailed(Response $response): ?bool
{
    // if psr response is json
    $contentType = $response->getPsrResponse()->getHeader('Content-Type');

    if (in_array('application/json', $contentType)) {
        return ! empty($response->json('errors'));
    }

    return null;
}
```

This PR would simplify it like this:
```php
public function hasRequestFailed(Response $response): ?bool
{
    if ($response->isJson()) {
        return ! empty($response->json('errors'));
    }

    return null;
}
```

I would argue, though, that I prefer a more strict comparison than `str_contains()` proposed in this PR.

mstfblci:
@SRWieZ Thank you for your support.

Sammyjo20 (changes requested):
Love this! Super simple as well. Would you be able to write two tests for these functions just to prove coverage? Many thanks

Sammyjo20 (approved):

Sammyjo20:
I'm going to have a Saloon day this weekend so I'll merge it then, thanks for the patience!

SRWieZ:
Glad to see you back! 

And thank you for all the work you put into Saloon.

PR #468, opened 2024-11-28
Fix | Resolve PHP 8.4 deprecation warnings

markwalet:
Resolve PHP 8.4 deprecation warnings by explicitly marking nullable arguments as nullable.

_Edit: It looks like the failing Actions are not specific for my PR, let me know if I need to fix something._


Sammyjo20 (approved):
Thank you for sorting this!

Sammyjo20 (approved):

PR #469, opened 2024-11-29
Feature | Add OAuth 1.0a Authenticator

fredbradley:
I was working on a project that needed authentication with OAuth1.0a. 

Here was the solution.


Sammyjo20:
Thanks @fredbradley I will take a look soon and see if I can potentially refactor this to be trait-style like the other OAuth2 implementations, unless you want to give it a go?

Sammyjo20:
Hey @fredbradley any further thoughts on moving to a trait?

fredbradley:
Hi @Sammyjo20, 

I wasn't sure there was need to make it into a trait. It's a one class Authenticator, just like the other Authenticators. 

Usage: 
```
protected function defaultAuth(): ?Authenticator
    {
        return new OAuth1Authenticator(
            consumerKey: config('services.xx.access_token'),
            consumerSecret: config('services.xx.secret'),
            token: config('services.xx.access_token'),
            tokenSecret: config('services.xx.token_secret')
        );
    }
```
I don't see the benefit of converting it to be trait like, this syntax is used elsewhere in Saloon. In `Http\Auth`

PR #477, opened 2025-01-17
Feature | Improve working with queries

aerni:
This PR provides a couple of improvements around working with queries:

- It adds the ability to define custom query builders
- It adds the ability to work with queries fluently

## Custom Query Builder

Defining a custom query builder works similar to defining a [custom response](https://docs.saloon.dev/the-basics/responses#custom-responses). Providing your own query builder allows you to add methods that mirror the supported parameters of the API you're working with. It's also super helpful for autocompletion in your code editor.

Custom query builders are simply extending the `Saloon\Repositories\ArrayStore` that is already being used for queries. This ensures a non-breaking change.

A simple example of a custom query builder:

```php
use Saloon\Repositories\ArrayStore;

class CloudflareStreamQueryBuilder extends ArrayStore
{
    public function limit(int $limit): self
    {
        $this->add('limit', $limit);

        return $this;
    }
}
```

How you'd register the query builder on a connector or request:

```php
class CloudflareStreamRequest extends Request
{
    protected string $queryBuilder = CloudflareStreamQueryBuilder::class;
}
```

How you'd use the query builder:

```php
$request->query()->limit(5)->add('foo', 'bar');
```

## Fluent Query Builder

A new `fluentQuery()` method makes working with queries a little smoother.

From this:

```php
$connector = new CloudflareConnector();

$connector->query()->limit(5)->add('foo', 'bar');
  
$connector->send($request);
```

To this:

```php
new CloudflareConnector()
    ->fluentQuery(fn ($query) => $query->limit(5)->add('foo', 'bar'))
    ->send($request);
```

I'd prefer to simplify things and merge the logic of the `fluentQuery()` method with the `query()` method. But that would require a method signature change and be a breaking change. I'll leave this decision up to you.

## Feature Request: Queries on resources

I'd love to make queries even more powerful by adding support for resources. This would be super powerful for building SDKs. Consider the following scenario, where you've got different resources that each need their own query builder class. Currently, you can only set the query builder on the connector, which sets the query builder for all resources. Or on each request directly, which can be a little tedious.

This is what I'd love to see:

```php
class CloudflareConnector extends Connector
{
    public function video(): VideoResource
    {
        return new VideoResource($this);
    }

    public function audio(): AudioResource
    {
        return new AudioResource($this);
    }
}
```

```php
class VideoResource extends BaseResource
{
    protected string $queryBuilder = VideoQueryBuilder::class;
}
```

```php
class AudioResource extends BaseResource
{
    protected string $queryBuilder = AudioQueryBuilder::class;
}
```

Then you can use it like this:

```php
new CloudflareConnector()
    ->video()
    ->fluentQuery(fn (VideoQueryBuilder $query) => $query->limit(5)->add('foo', 'bar'))
    ->all()
```

I hope this all makes sense. Thanks for your work on Saloon!

### DIFF
--- a/src/Traits/RequestProperties/HasQuery.php
+++ b/src/Traits/RequestProperties/HasQuery.php
@@ -4,8 +4,9 @@ declare(strict_types=1);
 
 namespace Saloon\Traits\RequestProperties;
 
-use Saloon\Repositories\ArrayStore;
+use Closure;
 use Saloon\Contracts\ArrayStore as ArrayStoreContract;
+use Saloon\Repositories\ArrayStore;
 
 trait HasQuery
 {
@@ -20,6 +21,16 @@ trait HasQuery
     public function query(): ArrayStoreContract
     {
         return $this->query ??= $this->resolveQueryBuilder()->set($this->defaultQuery());
+    }
+
+    /**
+     * Access the query parameters fluently
+     */
+    public function fluentQuery(?Closure $callback = null): self
+    {
+        $callback($this->query());
+
+        return $this;
     }
 
     /**
--- a/src/Traits/RequestProperties/HasQuery.php
+++ b/src/Traits/RequestProperties/HasQuery.php
@@ -11,6 +11,11 @@ use Saloon\Repositories\ArrayStore;
 trait HasQuery
 {
     /**
+     * Specify the default query builder class
+     */
+    protected string $queryBuilder = ArrayStore::class;
+
+    /**
      * Request Query Parameters
      */
     protected ArrayStoreContract $query;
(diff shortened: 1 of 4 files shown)


Sammyjo20:
I really appreciate the effort you have gone into this PR, thank you however  I don't feel this is quite right for Saloon. 

I hope you have a great day.

aerni:
Hi! Any particular reason why you feel this isn't quite right? Is it the implementation? Or the feature in general? I feel like the Custom Query Builder feature would be a great enhancement for building SDKs, providing a tight coupling to the respective API.

PR #480, opened 2025-02-06
Feature | Refactor merge logic for recursive and safe array merging

mfrieswyk:
Enhanced `merge` methods to support recursive merging with proper handling of associative and sequential arrays. Introduced helper methods to ensure safe merging and added tests to validate the behavior. This improves flexibility when working with complex nested data structures.

This allows a more flexible developer experience, whereby you can merge a modified body back into the request and all matching nodes will be updated while any new nodes are added.


mikezm:
This would be very helpful for some of the more complex payloads we're sending with Saloon now.

PR #492, opened 2025-04-27
Leverage PHPStan conditional return types for more precise return-type declarations

yankewei:


Sammyjo20 (approved):
Nice improvement, thank you!

PR #502, opened 2025-06-21
Fix | URLHelper incorrectly URLs with underscores

### LINKED ISSUE #498: Use `parse_url` instead of `FILTER_VALIDATE_URL` in `URLHelper` to allow underscores in domains
Domains including an underscore like: 

    https://example_site.laravel-sail.site:8080

Being used in a `SoloRequest` will result in a request being sent to:

    /https://example_site.laravel-sail.site:8080

This can be verified by adding the following test which will fail:

```php
// tests/Unit/URLHelperTest.php:16
['', 'https://example_domain.com', 'https://example_domain.com'],
```

```
Failed asserting that two strings are identical.
Expected :'https://example_domain.com'
Actual   :'/https://example_domain.com'
```

---

This is because the `FILTER_VALIDATE_URL` does not accept underscores in URLs even though it does work in domain names and is actively being used.

See [this PHP issue](https://github.com/php/php-src/issues/17842) about the topic and where it is also suggested to use `parse_url` instead.

Replacing the `isValidUrl` function to the below passas all tests including the new one above.

```php
/**
 * Check if the URL is a valid URL
 */
public static function isValidUrl(string $url): bool
{
    $parsed = parse_url($url);
    return array_key_exists('host', $parsed)
        && array_key_exists('scheme', $parsed);
}
```

Sammyjo20:
Fixes #498

### DIFF
--- a/src/Helpers/URLHelper.php
+++ b/src/Helpers/URLHelper.php
@@ -44,6 +44,11 @@ class URLHelper
      */
     public static function isValidUrl(string $url): bool
     {
+        // The following str_replace is used to get around an issue raised by PHP 8.4
+        // @see https://github.com/php/php-src/issues/17842
+
+        $url = str_replace('_', '-', $url);
+
         return ! empty(filter_var($url, FILTER_VALIDATE_URL));
     }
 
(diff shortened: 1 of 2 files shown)


Sammyjo20 (commented):
Self review

KSneijders:
@Sammyjo20 Nice and simple solution to the problem! Thanks!

PR #504, opened 2025-06-23
Add partial mocking

### LINKED ISSUE #501: Allow partial mocking
When testing end to end I might be hitting my own API which in turn sends to another API via background job or something. I don't have access to the connector or request(s) in these instances, and sometimes I would like to be able to mock some requests but not others.

Would be nice to be able to fall back to sending an actual request if a mock response for a given endpoint isn't found. Since mock responses are injected via `DetermineMockResponse` (a `RequestMiddleware`), I think you could do this just by returning the `$pendingRequest` instead of throwing a `NoMockResponseFoundException`.

epixian:
Adds an `allowFallback()` method to the `MockClient`. This returns the `$pendingRequest` instead of throwing `NoMockResponseFoundException`, allowing the request pipeline to continue (and send an actual request).

Closes #501.

### DIFF
--- a/src/Http/Faking/MockClient.php
+++ b/src/Http/Faking/MockClient.php
@@ -60,6 +60,13 @@ class MockClient
     protected static ?MockClient $globalMockClient = null;
 
     /**
+     * Whether to allow sending an actual request if no mock response is found.
+     *
+     * @var bool
+     */
+    protected bool $allowFallback = false;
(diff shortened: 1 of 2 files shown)


alexandre-tobia:
@Sammyjo20 Hi ! Any news on this ? Would be awesome to have this.

PR #515, opened 2025-09-10
Feature | Add baseUrl to OAuthConfig and Requests

kevinmade:
This PR introduces support for overriding the base URL at both the request level and within the `OAuthConfig`.

### DIFF
--- a/src/Helpers/OAuth2/OAuthConfig.php
+++ b/src/Helpers/OAuth2/OAuthConfig.php
@@ -32,6 +32,11 @@ class OAuthConfig
     protected string $redirectUri = '';
 
     /**
+     * The Base URL
+     */
+    protected ?string $baseUrl = null;
+
+    /**
      * The endpoint used for the authorization URL.
      */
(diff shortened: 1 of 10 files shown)


zngly-vlad:
showing support hoping this gets merged 👍

yankewei:
Looks like you just added a property to the OAuthConfig class just to hold a variable? That feels a little strange.

kevinmade:
Thats correct OAuthConfig is used for holding all OAuth-related configuration (credentials, endpoints, etc.). I added baseUrl because in the API that i want to use saloon for, the OAuth server uses a different base URL than the main API endpoints.

yankewei:
@kevinmade However, if the OAuth base URL differs from the main API's base URL, simply use the full URL when declaring the token endpoint in OAuthConfig. See an example in the official [documentation](https://docs.saloon.dev/digging-deeper/oauth2-authentication/oauth2-authentication#getting-started).

kevinmade:
Hadn’t seen that part of the documentation. Thanks for pointing it out.

PR #516, opened 2025-09-10
Add conditional return type to ClientCredentialsGrant getAccessToken

kevinmade:


craigpotter (approved):
Looks good although would be better with a PR description

src/Traits/OAuth2/ClientCredentialsGrant.php
```diff
@@ -27,6 +27,8 @@ trait ClientCredentialsGrant
      *
      * @param array<string> $scopes
      * @param callable(TRequest): (void)|null $requestModifier
+     *
+     * @return ($returnResponse is true ? \Saloon\Http\Response : \Saloon\Contracts\OAuthAuthenticator)
```
treyssatvincent, line 31: No need for Fully Qualified Domain Name as it's already used bellow in the signature:

```suggestion
     * @return ($returnResponse is true ? Response : OAuthAuthenticator)
```

treyssatvincent (approved):

src/Traits/OAuth2/ClientCredentialsGrant.php
kevinmade: I copied it from `Saloon\Traits\OAuth2\AuthorizationCodeGrant` to keep it consistent. I can change it if needed, but it was meant as a small, quick fix.

Sammyjo20 (approved):

Sammyjo20:
Thanks @kevinmade!

Issue #520, opened 2025-10-24
The PHP-CS-Fixer CI fix introduces breaking changes


yankewei:
The CI environment for php-cs-fixer uses PHP 8.4, so some string-related code has been updated to be compatible with this version—for example, replacing `rtrim(...)` with `mb_rtrim(...)`. Although the `mb_rtrim` function was introduced in PHP 8.4, our codebase remains fully compatible with PHP 8.1 and above.

yankewei:
Fixed in #525

PR #523, opened 2025-11-12
Fix: Use PHP 8.1 in CS Fixer CI workflow

yankewei:
## Problem
The previous CI configuration used `oskarstark/php-cs-fixer-ga` Docker image, which was running on
a newer PHP version (likely 8.4). This caused inconsistent code style fixes when compared to
running php-cs-fixer locally on PHP 8.1 (the project's minimum requirement).
## Solution
- Use `shivammathur/setup-php@v2` to explicitly set PHP 8.1
- Install dependencies via composer
- Run php-cs-fixer using the project's own dependency

### DIFF
--- a/.github/workflows/php-cs-fixer.yml
+++ b/.github/workflows/php-cs-fixer.yml
@@ -24,10 +24,15 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-      - name: Run PHP CS Fixer
-        uses: docker://oskarstark/php-cs-fixer-ga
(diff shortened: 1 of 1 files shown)


yankewei:
Fix #520

yankewei:
@Sammyjo20, Please take a look at this when you have a moment.

Sammyjo20:
Ah thanks for finding this @yankewei that's funny because I just found the same issue and have fixed in #525 :)

Sammyjo20:
All done in the latest release!

PR #536, opened 2026-03-06
Add PHPStan generics to Request, Response and send() for typed DTO inference

tomasvanrijsse:
Adds @template TDto to Request and Response, wires the generic through send() so that PHPStan and IDEs can infer the DTO type at the call site without any changes to runtime behaviour.

Users can now annotate their request classes:

    /** @extends Request<CustomerDto> */
    class CustomerRequest extends Request
    {
        public function createDtoFromResponse(Response $r): CustomerDto { ... }
    }

And the full chain resolves automatically:

    $response = $connector->send(new CustomerRequest()); // Response<CustomerDto>
    $dto      = $response->dto();                        // CustomerDto

All internal usages of Response and Request have been annotated with <mixed> as required by PHPStan's missingType.generics rule. No behaviour changes — pure docblock additions.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,7 +1,7 @@
 parameters:
 	ignoreErrors:
 		-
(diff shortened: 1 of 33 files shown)


lloricode:
1+

PR #542, opened 2026-03-17
Feature | v4 - Add opt in to base url overrides

JonPurvis:
Adds the ability to specify whether a Request can override the base URL set in the connector. You can also define it at the connector level.


src/Http/PendingRequest.php
```diff
@@ -316,4 +325,28 @@ protected function tap(callable $callable): static
...
         return $this;
     }
+
+    protected function resolveAllowBaseUrlOverrideForUrl(): bool
+    {
+        if ($this->request->allowBaseUrlOverride !== null) {
+            return $this->request->allowBaseUrlOverride;
+        }
+
+        if ($this->usesOAuthConfigTokenOrUserEndpoint() && method_exists($this->connector, 'oauthConfig')) {
```
Sammyjo20:
```suggestion
        if ($this->usesOAuthConfigTokenOrUserEndpoint() && method_exists($this->connector, 'oauthConfig') && $this->connector->oauthConfig instanceof OAuthConfig) {
```

Sammyjo20 (approved):

src/Http/PendingRequest.php
JonPurvis: Good call, this is done!

Sammyjo20 (approved):

PR #554, opened 2026-06-12
[4.x] Fix: Null header deprecated notice for PSR requests

pokyunn:
Notice: Since guzzlehttp/psr7 2.11: Passing null to MessageInterface::withHeader() is deprecated; guzzlehttp/psr7 3.0 requires string|string[].

I sugest this instead of removing the header, because its easier to debug when you see an empty reader, than no header set

### DIFF
--- a/tests/Feature/PsrTest.php
+++ b/tests/Feature/PsrTest.php
@@ -5,10 +5,10 @@ declare(strict_types=1);
 use GuzzleHttp\Psr7\Uri;
 use Saloon\Http\Faking\MockClient;
 use Saloon\Http\Faking\MockResponse;
-use Saloon\Tests\Fixtures\Requests\ModifiedPsrUserRequest;
-use Saloon\Tests\Fixtures\Requests\NullHeaderRequest;
-use Saloon\Tests\Fixtures\Connectors\ModifiedPsrRequestConnector;
 use Saloon\Tests\Fixtures\Connectors\TestConnector;
+use Saloon\Tests\Fixtures\Requests\NullHeaderRequest;
+use Saloon\Tests\Fixtures\Requests\ModifiedPsrUserRequest;
+use Saloon\Tests\Fixtures\Connectors\ModifiedPsrRequestConnector;
 
 test('the connector and request can modify the psr request when it is created', function () {
     $mockClient = new MockClient([
@@ -29,7 +29,7 @@ test('the connector and request can modify the psr request when it is created', 
     expect($response->getPsrRequest()->getHeaders())->toHaveKey('X-Howdy', ['Yeehaw']);
 });
 
-test('The psr request and readers must be converted to empty string', function () {
+test('The psr request headers must be converted to empty string', function () {
     $mockClient = new MockClient([
         MockResponse::make(headers: ['X-Null-Header' => null]),
     ]);
@@ -42,6 +42,6 @@ test('The psr request and readers must be converted to empty string', function (
     // The request will convert null to empty string
     expect($response->getPsrRequest()->getHeader('X-Null-Header')[0])->toBe('');
 
-    // The responde will convert null header to empty string
+    // The response will convert null header to empty string
     expect($response->headers()->get('X-Null-Header'))->toBe('');
 });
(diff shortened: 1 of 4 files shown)


JonPurvis (changes requested):
Hey @pokyunn 👋 

Thanks for the contribution. This makes sense. I believe the same change will need making here:

https://github.com/saloonphp/saloon/blob/v4/src/Http/Faking/FakeResponse.php#L127

Could you also add a test that covers this change too please?

Thanks

pokyunn:
Hi, thanks for reply

I'm not sure if its the best way to add the tests, can you check?

tests/Feature/PsrTest.php
```diff
@@ -26,3 +28,20 @@
 
     expect($response->getPsrRequest()->getHeaders())->toHaveKey('X-Howdy', ['Yeehaw']);
 });
+
+test('The psr request and readers must be converted to empty string', function () {
```
JonPurvis: typo, readers should be headers

tests/Feature/PsrTest.php
```diff
@@ -26,3 +28,20 @@
...
+
+    $connector = new TestConnector();
+    $connector->withMockClient($mockClient);
+
+    $response = $connector->send(new NullHeaderRequest());
+
+    // The request will convert null to empty string
+    expect($response->getPsrRequest()->getHeader('X-Null-Header')[0])->toBe('');
+
+    // The responde will convert null header to empty string
```
JonPurvis: typo, resonde should be response

tests/Fixtures/Requests/NullHeaderRequest.php
```diff
@@ -0,0 +1,40 @@
...
+            'X-Null-Header' => null,
+        ];
+    }
+
+    protected function defaultConfig(): array
+    {
+        return [
+            'timeout' => 5,
+        ];
+    }
```
JonPurvis: You can remove this, not needed.

JonPurvis (changes requested):
Thanks for making the changes. I've just spotted a couple of typos so if you could action the latest round of feedback, we're all good.

pokyunn:
Done.

Just for reference.

Emulated actions with act (host Fedora 44 > linuxbrew > act)
act -j tests --matrix os:ubuntu-latest --matrix php:8.5 --matrix stability:prefer-stable 
> Job succeeded

act -j phpstan
> Job succeeded

But cs-fixer fails, could not determinate why, maybe extensions in setup action
act -j lint
> Job failed: Your requirements could not be resolved to an installable set of packages.

pokyunn:
About cs-fixer action, Testing with ramsey/composer-install (same as phpstam) it works, but only after bumping php runtime to 8.3+ 

Reporting
```
...
[Code Style/lint]   ✅  Success - Main Run PHP CS Fixer [1.55928304s]
....
[Code Style/lint]   ❌  Failure - Main Commit Changes [1.081228059s]
....
[Code Style/lint]   ✅  Success - Post Cache Composer dependencies [156.903119ms]
[Code Style/lint]   ✅  Success - Post Install composer dependencies [252.473267ms]
[Code Style/lint] ⭐ Run Complete job
[Code Style/lint]   ✅  Success - Complete job
[Code Style/lint] 🏁  Job failed
```

Code to test (not commited as its not directly related to this code and will need php bump, so just reporting, maybe this should be an issue?)
```
.github/workflows/php-cs-fixer.yml

      - name: Setup PHP
        uses: shivammathur/setup-php@v2
        with:
          php-version: '8.3'
          coverage: none

      - name: Install composer dependencies
        uses: ramsey/composer-install@v2
```

JonPurvis:
I usually just run the cs fixer locally and commit that. I think there's an issue in this repo where it only runs it if it's Sam (creator of Saloon)

JonPurvis (approved):
👍

pokyunn:
Hi, any ETA?

I saw, the main is stale for a bit now

Sammyjo20 (approved):

Sammyjo20:
Thank you @pokyunn !

PR #555, opened 2026-06-15
Static analysis improvements for macros

bonroyage:
This PR improves the static analysis with PHPStan in projects when using macros. It is twofold:
1. Typehint `$this` inside the macro
2. Detect macros where used

For point 1, I took inspiration from Laravel's [Macroable trait](https://github.com/laravel/framework/blob/13.x/src/Illuminate/Macroable/Traits/Macroable.php#L27) and added `@param-closure-this`. This ensures that within the closure, `$this` refers to the class being macro-ed instead of the class the macro is registered (like a service provider).

For point 2, I took inspiration from Larastan and how they [detect macros](https://github.com/larastan/larastan/blob/3.x/src/Methods/MacroMethodsClassReflectionExtension.php). The class is largely a one-to-one copy with some Laravel specific code removed. For ease of use, I added an extension, so the developer can simply add an include to their phpstan.neon file. It also works with auto registering extensions if `phpstan/extension-installer` is used.

```
includes:
    - ./vendor/saloonphp/saloon/extension.neon
```

---

Carbon also supports macros, [but they do it slightly differently since they don't use a trait](https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/PHPStan/MacroExtension.php)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,5 +75,12 @@
         "pstan": [
             "./vendor/bin/phpstan analyse --memory-limit=1G"
         ]
+    },
+    "extra": {
+        "phpstan": {
+            "includes": [
+                "extension.neon"
+            ]
+        }
     }
 }
--- a/extension.neon
+++ b/extension.neon
@@ -1,0 +1,5 @@
+services:
+    -
+        class: Saloon\PHPStan\MacroMethodsClassReflectionExtension
+        tags:
+            - phpstan.broker.methodsClassReflectionExtension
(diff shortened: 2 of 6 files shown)


JonPurvis (approved):
This seems okay to me 👍 

We'll probably want to add something to docs regarding enabling the extension. The docs are on Gitbook though so @Sammyjo20 will need to sort that if he's happy with the rest of the PR.

PR #556, opened 2026-07-12
feat: add HTTP QUERY method support (RFC 10008)

A909M:
The HTTP `QUERY` method was recently published as [RFC 10008](https://www.rfc-editor.org/info/rfc10008/). It's a safe, idempotent method that carries its query in the request body instead of the URL, solving the limitations of GET for large or sensitive queries.

This PR adds `QUERY` to `Saloon\Enums\Method`.

```php
class SearchRequest extends Request
{
    protected Method $method = Method::QUERY;
}


JonPurvis (approved):
It makes sense to have this 👍

Sammyjo20 (approved):

PR #563, opened 2026-09-22
[4.x] assertSent with typehinting

gdebrauwer:
Currently, when you are faking multiple requests in a single test, you have to manually check which request you are receiving in the closure before making the real assertions on the request:
```php
Saloon::assertSent($request, $response) {
	return $request instance of FirstRequest && $request->userId === 1;
});

Saloon::assertSent($request, $response) {
	return $request instance of SecondRequest && $request->firstname === 'Taylor';
});
```
This PR simplifies that by allowing you to typehint the $request parameter:
```php
Saloon::assertSent(FirstRequest $request, $response) {
	return $request->userId === 1;
});

Saloon::assertSent(SecondRequest $request, $response) {
	return $request->firstname === 'Taylor';
});
```


gdebrauwer:
The tests that are failing are not related to the PR changes. They are curl errors, so I think the api that is being called in those tests is just failing.

